PR #241, opened 2021-11-03
Add warning to anonymization.md

yoid2000:
@cristianberneanu I realized that we need to describe the potentially dangerous usage scenarios in the documentation. This is not because there is any real danger of the app being misused, but to avoid some bad publicity whereby someone points out the bad cases before we do.

We should update the download files with these changes.


yoid2000:
Oh, I also wanted to fix the missing space in the sidebar help menu, but as near as I can tell it had already been fixed. Is this the case?

cristianberneanu:
> Oh, I also wanted to fix the missing space in the sidebar help menu, but as near as I can tell it had already been fixed. Is this the case?

Yes, it was fixed yesterday evening.

cristianberneanu (approved):

PR #245, opened 2021-11-04
Warn about nulls in the AID column

### LINKED ISSUE #190: Warn when the selected AID column contains empty values.
Empty values in the AID column might denote a possible data cleanup issue ...

pdobacz:
Closes #190 
This seems to work and be "by-the-book" according to my limited React knowledge, but please make sure that I'm not violating some important conventions or abstractions here.

Also I struggled to give things proper names, so let me know if any are off.

Effect:
image


src/types/index.ts
```diff
@@ -74,7 +74,12 @@ export type PreviewResponse = {
   rows: ResultRow[];
 };
 
-export type Response = LoadResponse | PreviewResponse;
+export type CheckNullAidsResponse = {
```
pdobacz, line 77: I think I missed this when cleaning names. I'll probably rename to `FindNullsResponse` for consistency

src/AidSelectionStep/AidSelectionStep.tsx
```diff
@@ -59,12 +59,7 @@ export const AidSelectionStep: FunctionComponent<AidSelectionProps> = ({ schema,
       </div>
       <div className="AidSelectionStep-reserved-space">
         {/* Render next step */}
-        {aidColumn && (
-          <>
-            <Divider />
-            {children({ aidColumn })}
-          </>
-        )}
+        {aidColumn && <>{children({ aidColumn })}</>}
```
cristianberneanu, line 62: I think it makes more sense to show the empty values warning here, since it is an effect of selecting an AID.
Then the divider can also be kept here, which is even nicer.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -123,6 +124,35 @@ function GeneralizationControls({
...
+            type="warning"
+            showIcon
+            icon={<InfoCircleOutlined />}
+            closable
+          />
+        </>
+      ) : null;
+    }
+
+    case 'in_progress':
```
cristianberneanu, line 150: maybe show a loading animation here?

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -123,6 +124,35 @@ function GeneralizationControls({
...
+    case 'failed':
+      return null;
```
cristianberneanu, line 152: and a generic error here, like "Something went wrong" for example (as it should almost never happen anyway).

src/types/index.ts
```diff
@@ -74,7 +74,12 @@ export type PreviewResponse = {
...
+  columns: ResultColumn[];
+  rows: ResultRow[];
```
cristianberneanu, line 79: This amount of data seems overkill for this call. I think a simple boolean field is enough here.

src/types/index.ts
cristianberneanu: `CheckAidResponse` or `HasEmptyValuesResponse` sound nicer to me.

anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -149,6 +149,25 @@ let handleExport
...
+    FindNullsRequest.AidColumn = aidColumn
+  }
+  =
+  let anonParams = AnonymizationParams.Default
+
+  let query =
+    $"""
+      SELECT %s{aidColumn}
+      FROM table
+      WHERE %s{aidColumn} is NULL or %s{aidColumn} = ''
```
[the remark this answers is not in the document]
cristianberneanu: Or `handleAidCheck` if you prefer it more.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -123,6 +124,35 @@ function GeneralizationControls({
...
+
+  switch (computedResult.state) {
+    case 'completed': {
+      return computedResult.value ? (
+        <>
+          <Alert
+            className="ColumnSelectionStep-notice"
+            message={
+              <>
+                <strong>CAUTION:</strong>The protected entity identifier column contains NULL or empty values.
```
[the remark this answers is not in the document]
edongashi: Or `missing values`

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -123,6 +124,35 @@ function GeneralizationControls({
...
 }
 
+function NullsInAidWarning(schema: TableSchema, aidColumn: string) {
+  const computedResult = useNullAid(schema, aidColumn);
+
+  switch (computedResult.state) {
+    case 'completed': {
+      return computedResult.value ? (
+        <>
+          <Alert
```
edongashi, line 134: The `<>` is needed if you have more than 1 child. For a single child you can remove it.

Also, this can be converted from ternary operator to `computedResult.value && <Alert ... />` because `false` values are ignored by react.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -164,6 +194,8 @@ export const ColumnSelectionStep: FunctionComponent<ColumnSelectionStepProps> =
 
   return (
     <>
+      {NullsInAidWarning(schema, aidColumn)}
```
edongashi, line 197: This should be `<NullsInAidWarning schema={schema} aidColumn={aidColumn} />` because it uses hooks and therefore must be treated like a component. Hook execution must always be deterministic.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -164,6 +194,8 @@ export const ColumnSelectionStep: FunctionComponent<ColumnSelectionStepProps> =
...
+      <Divider />
```
edongashi, line 198: Better to return this divider where it was like Cristian suggested.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
pdobacz: again - where should the error show up and what should it say, without freaking the user out unnecessarily? Given the circumstances and the secondary nature of this warning, I'd vote to not show anything, since 99,9% of the time the Aid column is fine.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
pdobacz: ok, I'll try to work it out or ask

src/ColumnSelectionStep/ColumnSelectionStep.tsx
pdobacz: I'll change this, but I'm not following the explanation or note on determinism. This is some subtle difference somewhere in this JSX thing, right?

src/types/index.ts
pdobacz: so I wanted to have a "has..." name only on the level, where we're returning a boolean, so I'll do `CheckAidResponse` if that's preferred

src/types/index.ts
pdobacz: hm, this is what we're returning from the `SELECT...` query underneath. I think that's beyond our control, or maybe I'm missing sth?

We're deriving the boolean result on a later stage of processing the JSON response.

anonymizer/OpenDiffix.Service/Program.fs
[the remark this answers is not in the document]
pdobacz: hm, are you sure though?

That feels like coupling the AID check with the particular data source underneath. If we support anything else than CSV, this will need to change.

Not that I want to future-proof things, but I'd prefer to keep this check generic and independent of the data source. WDYT?

(regardless of the above, agreed on the `handleAidCheck` name change, since we're more generic than just nulls :+1: )

src/ColumnSelectionStep/ColumnSelectionStep.tsx
edongashi: Hooks are identified by call order, meaning they should never be put in conditionals. In this case you trigger a hook while rendering. That's fine but it's hidden and future refactors can't know that there's a hook call inside the markup.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
pdobacz: for the latter part: actually, if I do that and https://github.com/diffix/desktop/pull/245/files#r743555212, then it complains about the `false` being returned in lieu of the component. It accepts `Element | null` only

src/AidSelectionStep/AidSelectionStep.tsx
```diff
@@ -57,6 +91,8 @@ export const AidSelectionStep: FunctionComponent<AidSelectionProps> = ({ schema,
           ))}
         </Select>
       </div>
+      <NullsInAidWarning schema={schema} aidColumn={aidColumn} />
+      <Divider />
```
edongashi, line 95: I would put the Divider further below so that it's rendered conditionally (as it is in current master):

```jsx
<div className="AidSelectionStep-reserved-space">
  {/* Render next step */}
  {aidColumn && (
    <>
      <Divider />
      {children({ aidColumn })}
    </>
  )}
</div>
```

anonymizer/OpenDiffix.Service/Program.fs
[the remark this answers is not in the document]
cristianberneanu: I don't expect we will add support for additional data sources anytime soon (or ever, likely).
But I don't feel strongly about it, leaving it as is is fine with me.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
cristianberneanu: OK, makes sense.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
cristianberneanu: I guess it's ok without it, since this check is optional anyway.

src/AidSelectionStep/AidSelectionStep.tsx
```diff
@@ -1,15 +1,49 @@
...
+  schema: TableSchema;
+  aidColumn: string;
+};
+
+const NullsInAidWarning: FunctionComponent<NullsInAidWarningProps> = ({ schema, aidColumn }) => {
+  const computedResult = useNullAid(schema, aidColumn);
+
+  switch (computedResult.state) {
+    case 'completed': {
+      return computedResult.value ? (
```
edongashi, line 25: The switch can be dropped and simplified to `return computedResult.state === 'completed' && computedResult.value ? ( ... ) : null`.

src/types/index.ts
cristianberneanu: But we don't have to return the output of the query to the GUI.
The service knows the semantics of the query, it should also be the one interpreting the result.
If, in the future, the check becomes more complicated, only the service has to be updated.

src/types/index.ts
```diff
@@ -74,12 +74,11 @@ export type PreviewResponse = {
   rows: ResultRow[];
 };
 
-export type CheckNullAidsResponse = {
-  columns: ResultColumn[];
-  rows: ResultRow[];
+export type HasMissingValuesResponse = {
+  hasMissingValues: boolean;
```
cristianberneanu, line 78: This feels a bit redundant. Calling the structure `CheckForMissingValuesResponse` avoids that, but is a bit verbose.

cristianberneanu (approved):

src/AidSelectionStep/AidSelectionStep.tsx
```diff
@@ -1,15 +1,41 @@
...
+  schema: TableSchema;
+  aidColumn: string;
+};
+
+const MissingAidWarning: FunctionComponent<MissingAidWarningProps> = ({ schema, aidColumn }) => {
+  const computedResult = useMissingAid(schema, aidColumn);
+
+  return computedResult.state === 'completed' && computedResult.value ? (
+    <Alert
+      className="ColumnSelectionStep-notice"
```
edongashi, line 25: Stale class name. `MissingAidWarning` or maybe not needed at all?

src/AidSelectionStep/use-missing-aid.ts
```diff
@@ -0,0 +1,41 @@
...
+export function useMissingAid(schema: TableSchema, aidColumn: string): ComputedData<boolean> {
+  const anonymizer = useAnonymizer();
+  const [result, setResult] = useState<ComputedData<boolean>>({ state: 'completed', value: false });
+
+  useEffect(() => {
+    if (!aidColumn) {
+      // For no AID we know immediately that there are no NULL values.
+      setResult({ state: 'completed', value: false });
+      return;
+    } else {
```
edongashi, line 14: Redundant else indentation because of early return

src/types/index.ts
pdobacz: alternative is to just return a plain `bool/boolean` (`true/false`). This gives up on the convention to keep all response/request structured. I coded it, it just works with 

```
type HasMissingValuesResponse = bool
```

and similar changes where needed. WDYT?

cristianberneanu: That sounds good too.

src/AidSelectionStep/AidSelectionStep.tsx
pdobacz: good catch, I'm still css-blind a bit. I'll reuse `"AidSelectionStep-notice"`, I guess style-wise these two notices don't differ

src/AidSelectionStep/use-missing-aid.ts
pdobacz: that's kinda deliberate... I've developed strong dislike towards early returns. But if that's too unidiomatic, I'll adjust

src/AidSelectionStep/AidSelectionStep.tsx
edongashi: We have this in-house CSS rule where we name classes based on components. If it's the root rendered element then it's just `ComponentName`. If it's some sub-element then we use `ComponentName-part`.

edongashi: The current component is `MissingAidWarning` so that should be the class name.

src/AidSelectionStep/use-missing-aid.ts
edongashi: Both `return` and `else` are redundant. Personally I'd go with less indentation, but feel free to go with either.

cristianberneanu (approved):

PR #246, opened 2021-11-11
Add OpenDiffix.Service.Test

### LINKED ISSUE #223: Add tests for the anonymization service.
Currently, we don't have any tests in this repository.
While the GUI code is harder to test automatically, the anonymization service code is a good candidate for that.
These tests should also run as part of the CI workflow.

pdobacz:
Closes #223 

Adds basic tests for each of 4 request types. Tests on a top level by supplying JSON requests and asserting on the structure of the results.

Relies on a modified CSV-enabled `generate.fsx`.

### DIFF
--- a/.github/workflows/dev_build.yml
+++ b/.github/workflows/dev_build.yml
@@ -20,12 +20,23 @@ jobs:
         run: npm install
       - name: Install Service dependencies
         working-directory: anonymizer
-        run: dotnet tool restore
+        run: |
+          dotnet tool restore
+          dotnet restore
       - name: Check files format
         run: npm run check-format
       - name: Run linter
         run: npm run lint
       - name: Build Service
         run: npm run build
+      - name: Generate test data
+        working-directory: anonymizer
+        run: dotnet fsi reference/data/generate.fsx
+      - name: Test Service
+        working-directory: anonymizer
+        run: dotnet test --no-restore --verbosity normal
+      - name: Check files format for Service
+        working-directory: anonymizer
+        run: dotnet fantomas --check -r .
       - name: Build App
         run: npm run package
--- a/.github/workflows/dev_build.yml
+++ b/.github/workflows/dev_build.yml
@@ -35,8 +35,5 @@ jobs:
       - name: Test Service
         working-directory: anonymizer
         run: dotnet test --no-restore --verbosity normal
-      - name: Check files format for Service
-        working-directory: anonymizer
-        run: dotnet fantomas --check -r .
       - name: Build App
         run: npm run package
--- a/anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
+++ b/anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
@@ -29,7 +29,7 @@ let ``Handles Load request`` () =
 let ``Handles HasMissingValues request`` () =
   let request =
     $"""
-    {{"type":"HasMissingValues","inputPath":"%s{dataPath}","aidColumn":"\"id\""}}
+    {{"type":"HasMissingValues","inputPath":"%s{dataPath}","aidColumn":"id"}}
     """
 
   [ "true"; "false" ] |> should contain (request |> mainCore)
@@ -38,7 +38,7 @@ let ``Handles HasMissingValues request`` () =
 let ``Handles Preview request`` () =
   let request =
     $"""
-    {{"type":"Preview","inputPath":"%s{dataPath}","aidColumn":"\"id\"","salt":"1","buckets":["\"age\"","\"city\""],"countInput":"Rows","rows":1000}}
+    {{"type":"Preview","inputPath":"%s{dataPath}","aidColumn":"id","salt":"1","buckets":["age","city"],"countInput":"Rows","rows":1000}}
     """
 
   let response = request |> mainCore
@@ -57,7 +57,7 @@ let ``Handles Export request`` () =
 
   let request =
     $"""
-    {{"type":"Export","inputPath":"%s{dataPath}","aidColumn":"\"id\"","salt":"1","buckets":["\"age\"","\"city\""],"countInput":"Rows","outputPath":"%s{outputFile.Path}"}}
+    {{"type":"Export","inputPath":"%s{dataPath}","aidColumn":"id","salt":"1","buckets":["age","city"],"countInput":"Rows","outputPath":"%s{outputFile.Path}"}}
     """
 
   request |> mainCore |> should equal ""
--- a/anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
+++ b/anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
@@ -5,79 +5,62 @@ open FsUnit.Xunit
 
 open OpenDiffix.Service.Program
 
-let private dataDirectory = __SOURCE_DIRECTORY__ + "/../../data/data.sqlite"
+let private dataPath = __SOURCE_DIRECTORY__ + "/../reference/data/customers.csv"
 
-// [<Fact>]
-// let ``Prints version`` () =
-//   [| "--version" |] |> mainCore |> should not' (be Empty)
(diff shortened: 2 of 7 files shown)


anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -149,6 +148,8 @@ let handleExport
 
   File.WriteAllText(outputPath, output)
 
+  None
```
pdobacz, line 151: if anyone is wondering - why use `Option` here? Here's a full story:

After a naive refactor, whereby I just returned "" here, which was later `printfn`ed, an obscure problem arose on the GUI end - file export would work but would show a popup about failed export.

The reason was that the JSON response being returned from here was `'\n'`, which isn't valid JSON from JSON.parse perspective, at the same time being a truthy JS value, so `json ? JSON.parse(json)` threw errors.

cristianberneanu: Couldn't that have been solved easier by using `printf` or `Console.Out.Write` instead of `printfn`?

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
```diff
@@ -5,79 +5,62 @@ open FsUnit.Xunit
...
-// [<Fact>]
-// let ``Counts in non-anonymized tables`` () =
-//   [| "-f"; dataDirectory; "-q"; "SELECT count(*) FROM purchases" |]
-//   |> mainCore
-//   |> should not' (be Empty)
+[<Fact>]
+let ``Handles Load request`` () =
+  let request =
+    $"""
+    {{"type":"Load","inputPath":"%s{dataPath}","rows":10000}}
```
cristianberneanu, line 21: Why are the double curly brackets needed?

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
```diff
@@ -5,79 +5,62 @@ open FsUnit.Xunit
...
-// [<Fact>]
-// let ``Executes example batch query`` () =
-//   [| "--queries-path"; __SOURCE_DIRECTORY__ + "/../../queries-sample.json" |]
-//   |> mainCore
-//   |> should not' (be Empty)
+[<Fact>]
+let ``Handles Preview request`` () =
+  let request =
+    $"""
+    {{"type":"Preview","inputPath":"%s{dataPath}","aidColumn":"\"id\"","salt":"1","buckets":["\"age\"","\"city\""],"countInput":"Rows","rows":1000}}
```
cristianberneanu, line 41: It would be a bit more readable to instead create the input as a map and then jsonify it.

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
```diff
@@ -5,79 +5,62 @@ open FsUnit.Xunit
...
+
+  let request =
+    $"""
+    {{"type":"Export","inputPath":"%s{dataPath}","aidColumn":"\"id\"","salt":"1","buckets":["\"age\"","\"city\""],"countInput":"Rows","outputPath":"%s{outputFile.Path}"}}
+    """
+
+  request |> mainCore |> should equal Option.None
+
+  System.IO.File.ReadAllLines(outputFile.Path)
+  |> should contain "\"age\",\"city\",\"diffix_count\""
```
cristianberneanu, line 66: Asserting the line count would also be good here.

.github/workflows/dev_build.yml
```diff
@@ -20,12 +20,23 @@ jobs:
...
         run: npm run build
+      - name: Generate test data
+        working-directory: anonymizer
+        run: dotnet fsi reference/data/generate.fsx
+      - name: Test Service
+        working-directory: anonymizer
+        run: dotnet test --no-restore --verbosity normal
+      - name: Check files format for Service
+        working-directory: anonymizer
+        run: dotnet fantomas --check -r .
```
cristianberneanu, line 40: This is already done in the `check files format` section. The tests project should be included in the `npm` command.

cristianberneanu (approved):

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
pdobacz: this is the way to "escape" `{}` in a `$""""""` literal

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
pdobacz: I considered this but rejected the idea b/c:
1/ it's now super-easy to just copy paste these requests from manual testing logs
2/ JSON seemed to require significant boilerplate, judging by the `JSONDecodersEncoders.fs` work I did recently, but maybe if I use `Map` here (which is what you suggest, right), it's not that bad?

I'll give it another thought

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
edongashi: Probably to print a literal curly instead of interpolation delimiter.

anonymizer/OpenDiffix.Service/Program.fs
pdobacz: no strong opinion here, the newline at the end seemed like a nice touch, in case anyone would like to fire these requests manually.

Let's change

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
cristianberneanu: Thanks, I was under the impression interpolation happens only after a `%` char.
But it seems the format specifier is optional...

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
pdobacz: hm, to me both alternatives are worse off

```
  let request =
    $"""
    {{"type":"HasMissingValues","inputPath":"%s{dataPath}","aidColumn":"\"id\""}}
    """

  // vs
  let requestMap = Map [ ("type", "HasMissingValues"); ("inputPath", dataPath); ("aidColumn", "id") ]
  let request = Encode.Auto.toString (2, requestMap)

  // vs
  let requestMap = {| Type = "HasMissingValues"; InputPath = dataPath; AidColumn = "id"|}
  let request = Encode.Auto.toString (2, requestMap, caseStrategy = CamelCase)
```

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
cristianberneanu: Also, the export behavior seems wrong here: the last column should be called "count".

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
pdobacz: oh, and I noticed that the escaping is redundant in the string version 7b1bc2b7

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
pdobacz: > Also, the export behavior seems wrong here: the last column should be called "count".

let's tackle this separately

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
cristianberneanu: Personally, I like the anon type version more. But I won't insist on it.

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
pdobacz: oh this is simple. let's just ship it cd84a677

cristianberneanu (approved):

PR #253, opened 2021-11-15
Add a tooltip for numeric generalization in Preview step

### LINKED ISSUE #152: Show range as tooltip in cells of generalized columns
Hovering over a generalized numeric column should display the range as a tooltip. For example if the table cell shows the value `20` and is generalized with a bin size of `10`, then its hover tooltip would be `20-30`.

pdobacz:
Closes #152 

It's been a bit difficult to solve this, respecting all separations of concerns requirements. It came out a bit over-functional perhaps, lemme know if you can follow through.

Also, there's an issue with the *styling* of the tooltips. We have 3 options:
1. continue polishing so that it's super consistent across the app
2. be somewhat consistent
3. be noticeably inconsistent, but simplify code

So far I've chosen the (2.). Let's see if you can spot the inconsistency and/or find the code too complex and we'll see, if we need to switch to either 1 or 3. I think current choice of (2.) is "good enough".

### DIFF
--- a/src/AnonymizationStep/AnonymizedResultsTable.tsx
+++ b/src/AnonymizationStep/AnonymizedResultsTable.tsx
@@ -1,4 +1,5 @@
 import React, { FunctionComponent, useState } from 'react';
+import { Tooltip } from 'antd';
 
 import { columnSorter, formatPercentage, relativeNoise, ResponsiveTable } from '../shared';
 import {
@@ -9,6 +10,7 @@ import {
   DisplayMode,
   RowData,
   Value,
+  TooltipFunction,
 } from '../types';
 import { DisplayModeSwitch } from './DisplayModeSwitch';
 
@@ -21,34 +23,60 @@ type TableRowData = RowData & {
 
 // Columns
 
+function cellTrivialTooltip(displayValue: string) {
+  return <Tooltip title={displayValue}>{displayValue}</Tooltip>;
+}
+
+function cellNullTooltip() {
+  return (
+    <Tooltip title="NULL">
+      <i>NULL</i>
+    </Tooltip>
+  );
+}
+
+function cellTooltip(tooltip: string, displayValue: string) {
+  return <Tooltip title={tooltip}>{displayValue}</Tooltip>;
+}
+
 function renderValue(v: Value) {
+  // we must disable the default `ellipsis: {showTitle: true}` tooltip, so we're serving
+  // a trivial Tooltip node everywhere
   if (v === null) {
-    return <i>NULL</i>;
+    return cellNullTooltip();
   } else {
-    return v.toString();
+    return cellTrivialTooltip(v.toString());
   }
 }
 
+function buildRenderTooltipValue(valueToTooltip: TooltipFunction) {
+  // see note on showTitle above
+  return (v: Value) => {
+    if (v === null) {
+      return cellNullTooltip();
+    } else {
+      return cellTooltip(valueToTooltip(v), v.toString());
+    }
+  };
+}
+
 function renderLowCountValue(v: Value) {
-  return v === null ? '-' : v.toString();
+  // see note on showTitle above
+  return cellTrivialTooltip(v === null ? '-' : v.toString());
 }
 
 function renderRelativeNoiseValue(v: Value) {
-  return v === null ? '-' : formatPercentage(v as number);
+  // see note on showTitle above
+  return cellTrivialTooltip(v === null ? '-' : formatPercentage(v as number));
 }
 
-function makeColumnData(
-  title: string,
-  dataIndex: string,
-  type: ColumnType,
-  render: (v: Value) => React.ReactNode = renderValue,
-) {
+function makeColumnData(title: string, dataIndex: string, type: ColumnType, render: (v: Value) => React.ReactNode) {
   return {
     title,
     dataIndex,
     render,
     sorter: columnSorter(type, dataIndex),
-    ellipsis: true,
+    ellipsis: { showTitle: false },
   };
 }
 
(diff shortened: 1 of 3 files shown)


src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -21,34 +23,60 @@ type TableRowData = RowData & {
...
+    </Tooltip>
+  );
+}
+
+function cellTooltip(tooltip: string, displayValue: string) {
+  return <Tooltip title={tooltip}>{displayValue}</Tooltip>;
+}
+
 function renderValue(v: Value) {
+  // we must disable the default `ellipsis: {showTitle: true}` tooltip, so we're serving
```
cristianberneanu, line 43: I don't understand this comment, since `showTitle` is set to `false` later on.

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -58,17 +86,22 @@ const mapColumn = (mode: DisplayMode) => (column: AnonymizedResultColumn, i: num
...
-          makeColumnData(column.name + ' (original)', i + '_real', AGG_COLUMN_TYPE),
+          makeColumnData(column.name + ' (original)', i + '_real', AGG_COLUMN_TYPE, renderValue),
           makeColumnData('Distortion', i + '_diff', 'real', renderRelativeNoiseValue),
         ];
     }
   }
 
-  return [makeColumnData(column.name, i.toString(), column.type)];
+  if (column.type === 'integer' || column.type === 'real') {
+    let renderTooltipValue = buildRenderTooltipValue(column.valueToTooltip);
```
cristianberneanu, line 100: Please fix this warning.

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -58,17 +86,22 @@ const mapColumn = (mode: DisplayMode) => (column: AnonymizedResultColumn, i: num
...
+    return [makeColumnData(column.name, i.toString(), column.type, renderTooltipValue)];
```
cristianberneanu, line 101: I think the call to `makeColumnData` can be extracted out.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -49,11 +51,19 @@ function minBinSize(columnType: ColumnType) {
   return columnType === 'real' ? MIN_BIN_SIZE_REAL : MIN_BIN_SIZE_INTEGER;
 }
 
+function numericGeneralizationTooltip(binSize: number, binStart: number) {
+  return binStart.toString() + '-' + (binStart + binSize).toString();
```
cristianberneanu, line 55: It is not clear from this that the upper bound is non-inclusive. Maybe `[0; 1)` is better?

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -49,11 +51,19 @@ function minBinSize(columnType: ColumnType) {
   return columnType === 'real' ? MIN_BIN_SIZE_REAL : MIN_BIN_SIZE_INTEGER;
 }
 
+function numericGeneralizationTooltip(binSize: number, binStart: number) {
```
cristianberneanu, line 54: The order of the arguments is a bit weird. Usually, it is the other way around.

src/types/index.ts
```diff
@@ -85,8 +89,11 @@ export type ResultColumn = {
 
 export type ResultRow = Value[];
 
+export type NonNullValue = boolean | number | string;
 export type Value = boolean | number | string | null;
```
cristianberneanu, line 93:
```ts
export type Value = boolean | number | string;
export type NullableValue = Value | null;
```
would make more sense.

src/types/index.ts
```diff
@@ -85,8 +89,11 @@ export type ResultColumn = {
...
 
+export type TooltipFunction = (v: NonNullValue) => string;
```
cristianberneanu, line 95: Why do you need to accept `boolean | number | string` here instead of just `number`?

src/types/index.ts
```diff
@@ -95,12 +102,18 @@ export type AnonymizedQueryResult = {
...
+      name: string;
+      type: TooltipColumnType;
+      valueToTooltip: TooltipFunction;
+    }
+  | {
+      name: string;
+      type: AnonymizedColumnType;
+    };
+
+export type AnonymizedColumnType = NoTooltipColumnType | 'aggregate';
```
cristianberneanu, line 116: I don't like that tooltip information pollutes the type system.
I think it would be better to inline the types here or allow for empty `valueToTooltip` callbacks.

src/types/index.ts
```diff
@@ -95,12 +102,18 @@ export type AnonymizedQueryResult = {
...
-  name: string;
-  type: AnonymizedColumnType;
-};
-
-export type AnonymizedColumnType = ColumnType | 'aggregate';
+export type AnonymizedResultColumn =
+  | {
+      name: string;
+      type: TooltipColumnType;
+      valueToTooltip: TooltipFunction;
```
cristianberneanu, line 109: maybe `tooltipRenderer` or `tooltipGenerator`?

src/AnonymizationStep/AnonymizedResultsTable.tsx
cristianberneanu: The way the comment was phrased, I understood that we're serving tooltips everywhere in order to disable the default behavior, which is exactly the inverse of what is happening. Maybe rephrase it like this:
"Because we later disable default tooltips, we must serve a trivial `Tooltip` node everywhere."

src/AnonymizationStep/AnonymizedResultsTable.tsx
cristianberneanu: No strong opinion from me either, it can stay like this.

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: yup, this is more or less the vein of the new wording. Thx!

src/types/index.ts
cristianberneanu: True, nothing is being rendered yet, but something is being created, made or formatted.

pdobacz: I think it's down to `tooltipCallback` or `tooltipMaker` then. 

(`formatter` is too weak, since there's logic and data (`binSize`) involved.).

Actually... the difficulty in picking a name here convinces me this smells of bad abstraction. I'll spend a second to come up with an alternative to this callback model.

src/types/index.ts
edongashi: > I don't like that tooltip information pollutes the type system.

I agree. Can the tooltip specific types be exported from the table component file?

src/types/index.ts
pdobacz: see https://github.com/diffix/desktop/pull/255

pdobacz:
Closing in favor of #255

PR #255, opened 2021-11-16
Add a tooltip for numeric generalization in Preview step - without callbacks

### LINKED ISSUE #152: Show range as tooltip in cells of generalized columns

pdobacz:
This is a simplified version of #253 . Closes #152 

From my own gut feel and the comments I gathered that the callback model isn't working out. This version keeps everything to the `AnonymizedResultTable` component, which only needs to be aware of the `bucketColumns` context from previous step, which I think is cleaner than the type-mess that the previous version caused.

Sorry about two reviews, I hope this is now much easier to review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,7 @@
     "package": "electron-forge package",
     "make": "electron-forge make",
     "publish": "electron-forge publish",
-    "lint": "eslint --ext .ts,.tsx .",
-    "lint-ci": "eslint --max-warnings 0 --ext .ts,.tsx .",
+    "lint": "eslint --max-warnings 0 --ext .ts,.tsx .",
     "check-format": "prettier --check src && cd anonymizer && dotnet fantomas --check -r .",
     "format": "prettier --write src && cd anonymizer && dotnet fantomas -r ."
   },
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "make": "electron-forge make",
     "publish": "electron-forge publish",
     "lint": "eslint --ext .ts,.tsx .",
+    "lint-ci": "eslint --max-warnings 0 --ext .ts,.tsx .",
     "check-format": "prettier --check src && cd anonymizer && dotnet fantomas --check -r .",
     "format": "prettier --write src && cd anonymizer && dotnet fantomas -r ."
   },
--- a/src/AnonymizationStep/AnonymizedResultsTable.tsx
+++ b/src/AnonymizationStep/AnonymizedResultsTable.tsx
@@ -23,11 +23,14 @@ type TableRowData = RowData & {
 
 // Columns
 
-function cellDefaultTooltip(displayValue: string) {
+// with `ellipsis: { showTitle: false }` set for the column, we're disabling the default tooltip
+// so that the styling of tooltips is more consistent. This forces us to provide Tooltip in all cases
+// even if the tooltip shown is same as the valueCell value.
+function plainStringCell(displayValue: string) {
   return <Tooltip title={displayValue}>{displayValue}</Tooltip>;
 }
 
-function cellNullTooltip() {
+function nullCell() {
   return (
     <Tooltip title="NULL">
       <i>NULL</i>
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "make": "electron-forge make",
     "publish": "electron-forge publish",
     "lint": "eslint --ext .ts,.tsx .",
+    "lint-ci": "eslint --max-warnings 0 --ext .ts,.tsx .",
```
edongashi, line 18: What does this do?

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -21,54 +23,97 @@ type TableRowData = RowData & {
...
+function cellNullTooltip() {
+  return (
+    <Tooltip title="NULL">
+      <i>NULL</i>
+    </Tooltip>
+  );
+}
+
+function cellNumericGeneralizedTooltip(binSize: number, v: number) {
+  const displayValue = v.toString();
```
edongashi, line 39: Not required to be converted to string. `<Tooltip title={tooltip}>{v}</Tooltip>` should work?

package.json
pdobacz: it's used in the CI [here](https://github.com/diffix/desktop/pull/255/files#diff-2ace50007d05636bdbb4ee32251052be22a3bcb9f01108e2cbb7a2df05abc36eR29) to catch missed linting warnings like [this](https://github.com/diffix/desktop/pull/253#discussion_r749469104)

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -21,54 +23,97 @@ type TableRowData = RowData & {
...
-  render: (v: Value) => React.ReactNode = renderValue,
-) {
+function makeColumnData(title: string, dataIndex: string, type: ColumnType, render: (v: Value) => React.ReactNode) {
   return {
     title,
     dataIndex,
     render,
     sorter: columnSorter(type, dataIndex),
-    ellipsis: true,
+    ellipsis: { showTitle: false },
```
edongashi, line 96: Do we lose tooltip in column titles now?

package.json
cristianberneanu: I don't think a separate command is warranted here. The original one should treat warnings as errors.

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -21,54 +23,97 @@ type TableRowData = RowData & {
...
+  const tooltip = `[${v}; ${v + binSize})`;
+  return <Tooltip title={tooltip}>{displayValue}</Tooltip>;
+}
+
+function cellTooltip(bucketColumn: BucketColumn | undefined, v: number | boolean | string) {
+  if (
+    bucketColumn &&
+    (bucketColumn.type === 'integer' || bucketColumn.type === 'real') &&
+    bucketColumn.generalization &&
+    bucketColumn.generalization.binSize
```
cristianberneanu, line 49: This condition seems overly defensive. When can the `bucketColumn` be `undefined` or `binSize` field be missing?

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -21,54 +23,97 @@ type TableRowData = RowData & {
...
+  return <Tooltip title={displayValue}>{displayValue}</Tooltip>;
+}
+
+function cellNullTooltip() {
+  return (
+    <Tooltip title="NULL">
+      <i>NULL</i>
+    </Tooltip>
+  );
+}
```
cristianberneanu, line 36: I don't think it makes a lot of sense to put `tooltip` in these functions' names.
The fact that they add a tooltip is not that meaningful. What is important is that they define a cell's content. They should be called just `nullCell | stringCell | rangeCell | ...`.

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: nope. If this was left at `true` we'd have 2 tooltips for the columns that have the generalization (and with different style). 

At current state, we have always a single tooltip, with style consistent per table. See comment in the [original PR](https://github.com/diffix/desktop/pull/253#issue-780588667)

cristianberneanu (dismissed):
Yes, this feels simpler  👍

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: Oh, you've edited your comment. So I think this `showTitle` setting has a confusing name. It's about tooltips in the `ellipsis`ed cells

package.json
pdobacz: if this doesn't break anyones existing workflow/integration, I agree to have just the strict `lint`

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: hm, I think I inherited from the code in the other module, where it was indeed nullable, good catch!

cristianberneanu (approved):

PR #256, opened 2021-11-16
Add a check for updates on app startup

### LINKED ISSUE #252: Notify about new versions.
If a new version is available, a notification should be shown on startup.
We first need to store the current version in the app.

pdobacz:
Closes #252 

I also tried to make the updates message clickable, with a URL pointing to the releases page, but this somehow isn't trivial. Given small returns from such an addition, I gave up and I'm just showing a plain message.

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -9,6 +9,7 @@ import { Docs, DocsFunctionsContext, PageId } from '../Docs';
 import { Notebook } from '../Notebook';
 
 import './App.css';
+import { useCheckUpdates } from './use-check-updates';
 
 const { TabPane } = Tabs;
 
@@ -70,6 +71,8 @@ function setWindowTitle(state: AppState) {
 
 export const App: FunctionComponent = () => {
   const [state, updateState] = useImmer(initialAppState);
+
+  useCheckUpdates();
 
   function onEdit(targetKey: unknown, action: 'add' | 'remove'): void {
     switch (action) {
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -7,9 +7,9 @@ import { find, findIndex } from 'lodash';
 import { AnonymizerContext, anonymizer, useStaticValue } from '../shared';
 import { Docs, DocsFunctionsContext, PageId } from '../Docs';
 import { Notebook } from '../Notebook';
+import { useCheckUpdates } from './use-check-updates';
 
 import './App.css';
-import { useCheckUpdates } from './use-check-updates';
 
 const { TabPane } = Tabs;
 
--- a/src/App/use-check-updates.ts
+++ b/src/App/use-check-updates.ts
@@ -6,7 +6,7 @@ async function checkForUpdates() {
     const newerVersion = await window.checkForUpdates();
     if (newerVersion) {
       message.info({
-        content: `New version ${newerVersion} available! See Help -> Latest Releases`,
+        content: `New version ${newerVersion} available! See Help ► Latest Releases`,
         key: 'useCheckUpdates',
         duration: 10,
       });
--- a/src/App/use-check-updates.ts
+++ b/src/App/use-check-updates.ts
@@ -1,0 +1,27 @@
+import { message } from 'antd';
+import { useEffect, useState } from 'react';
+
+export function useCheckUpdates(): void {
+  const [, setCheckedUpdates] = useState(false);
+  useEffect(() => {
+    message.loading({ content: `Checking for updates...`, key: 'useCheckUpdates', duration: 0 });
+    try {
+      setCheckedUpdates(true);
+      window.checkForUpdates().then((newerVersion) => {
+        if (newerVersion) {
+          message.info({
+            content: `New version ${newerVersion} available! See Help -> Latest Releases`,
+            key: 'useCheckUpdates',
+            duration: 10,
+          });
+        } else {
+          message.success({ content: 'No updates available!', key: 'useCheckUpdates', duration: 2 });
+        }
+      });
+    } catch (e) {
+      console.error(e);
+      message.error({ content: 'Checking for updates failed!', key: 'useCheckUpdates', duration: 2 });
+    }
+    return () => {};
+  }, []);
+}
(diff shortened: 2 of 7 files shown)


src/App/App.tsx
```diff
@@ -9,6 +9,7 @@ import { Docs, DocsFunctionsContext, PageId } from '../Docs';
 import { Notebook } from '../Notebook';
 
 import './App.css';
+import { useCheckUpdates } from './use-check-updates';
```
edongashi, line 12: Import should be before CSS with other relative imports (line 10).

src/App/use-check-updates.ts
```diff
@@ -0,0 +1,27 @@
...
+          message.info({
+            content: `New version ${newerVersion} available! See Help -> Latest Releases`,
+            key: 'useCheckUpdates',
+            duration: 10,
+          });
+        } else {
+          message.success({ content: 'No updates available!', key: 'useCheckUpdates', duration: 2 });
+        }
+      });
+    } catch (e) {
```
edongashi, line 21: This catch will not work because the function is not async

src/App/use-check-updates.ts
```diff
@@ -0,0 +1,27 @@
...
+      setCheckedUpdates(true);
+      window.checkForUpdates().then((newerVersion) => {
+        if (newerVersion) {
+          message.info({
+            content: `New version ${newerVersion} available! See Help -> Latest Releases`,
+            key: 'useCheckUpdates',
+            duration: 10,
+          });
+        } else {
+          message.success({ content: 'No updates available!', key: 'useCheckUpdates', duration: 2 });
```
edongashi, line 18: We shouldn't say anything if there are no updates.

src/main.ts
```diff
@@ -233,3 +236,13 @@ ipcMain.handle('set_main_window_title', (_event, title: string) => {
   const mainWindow = BrowserWindow.getAllWindows()[0];
   mainWindow?.setTitle(title);
 });
+
+ipcMain.handle('check_for_updates', async (_event) => {
+  const response = await fetch('https://api.github.com/repos/diffix/desktop/releases?per_page=1');
```
edongashi, line 241: We should ignore prereleases from notifying the user.

src/App/use-check-updates.ts
```diff
@@ -0,0 +1,27 @@
...
+        }
+      });
+    } catch (e) {
+      console.error(e);
+      message.error({ content: 'Checking for updates failed!', key: 'useCheckUpdates', duration: 2 });
+    }
+    return () => {};
```
edongashi, line 25: Not required to return this if it's no-op.

src/App/use-check-updates.ts
```diff
@@ -0,0 +1,27 @@
+import { message } from 'antd';
+import { useEffect, useState } from 'react';
+
+export function useCheckUpdates(): void {
+  const [, setCheckedUpdates] = useState(false);
```
edongashi, line 5: Why do we need this state? I think it's not required. The empty dependencies `[]` in the `useEffect` means that it will be called only once during app mount.

src/App/use-check-updates.ts
```diff
@@ -0,0 +1,27 @@
...
+  useEffect(() => {
+    message.loading({ content: `Checking for updates...`, key: 'useCheckUpdates', duration: 0 });
+    try {
+      setCheckedUpdates(true);
+      window.checkForUpdates().then((newerVersion) => {
+        if (newerVersion) {
+          message.info({
+            content: `New version ${newerVersion} available! See Help -> Latest Releases`,
```
edongashi, line 13: Can we put a clickable link here?

src/main.ts
pdobacz: settled on:

> edon  15 minutes ago
> so to summarize: fetch /latest, if 404 return null, else compare versions
> piotr  14 minutes ago
>  \+ OK to crash on anything else, as long as the app lives

src/App/use-check-updates.ts
pdobacz: > I also tried to make the updates message clickable, with a URL pointing to the releases page, but this somehow isn't trivial. Given small returns from such an addition, I gave up and I'm just showing a plain message.

edongashi: From ant docs I see that `content` can be a ReactNode. Meaning `conent: <>Click <a href="...">here</a> for latest version</>` should work.

edongashi (dismissed):

src/App/use-check-updates.ts
```diff
@@ -0,0 +1,24 @@
+import { message } from 'antd';
+import { useEffect } from 'react';
+
+async function checkForUpdates() {
+  try {
+    const newerVersion = await window.checkForUpdates();
+    if (newerVersion) {
+      message.info({
+        content: `New version ${newerVersion} available! See Help -> Latest Releases`,
```
edongashi, line 9: The "programmer's arrow" is not something I'd present to a regular user.

pdobacz: suggestions? drop altogether? nothing comes to mind :/

src/App/use-check-updates.ts
pdobacz: I tried multiple ways to produce this ReactNode and make whatever it has clickable (and pointing to an external browser) but failed miserably.

What you suggest also doesn't work :(, it shows the HTML tag verbatim

src/App/use-check-updates.ts
edongashi: If we put a link to releases then we can remove this altogether.

pdobacz: `See Help ► Latest Releases` maybe?

src/App/use-check-updates.ts
edongashi: Interesting... Nevermind then

src/App/use-check-updates.ts
pdobacz: supposedly `>` is a standard, I found a thread citing https://support.apple.com/en-us/HT201251#authorize

edongashi: Works for me. There are also one-character arrows, or a slash etc.

edongashi (approved):

PR #257, opened 2021-11-16
Inject simple LED hook in query pipeline

edongashi:
The implementation is different from Paul's pseudocode in order to reduce complexity from `O(n_lc_buckets * n_cols * n_buckets)` to roughly `O(n_lc_buckets * 1 * n_buckets)`

### DIFF
--- a/anonymizer/OpenDiffix.Service/Led.fs
+++ b/anonymizer/OpenDiffix.Service/Led.fs
@@ -92,8 +92,8 @@ let private executeLed executionContext (childPlan, groupingLabels, aggregators)
 
   // Group rows into buckets and keep track of low-count/count arguments
   for row in Executor.execute executionContext childPlan do
-    let group = groupingLabels |> Array.map (Expression.evaluate row)
     let argEvaluator = Expression.evaluate row
+    let group = groupingLabels |> Array.map argEvaluator
 
     let bucket =
       match state.TryGetValue(group) with
@@ -136,6 +136,19 @@ let private executeLed executionContext (childPlan, groupingLabels, aggregators)
     |> Seq.toArray
 
   // Merge victim buckets into sibling buckets
+  //
+  // for each low count bucket:
+  //   keep track of siblings per column (initially all empty lists)
+  //
+  //   for each other bucket:
+  //     find index where a single column is different
+  //     if such index:
+  //       add bucket to that column's siblings
+  //
+  //   if column siblings contains an empty list:
+  //     for each list of siblings per column:
+  //       if siblings.length = 1 and not siblings[0].lowCount:
+  //         merge bucket into siblings[0]
   buckets
   |> Array.iteri (fun i victimBucket ->
     if victimBucket.IsLowCount then
--- a/anonymizer/OpenDiffix.Service/Led.fs
+++ b/anonymizer/OpenDiffix.Service/Led.fs
@@ -8,8 +8,8 @@ let private findSingleIndex cond arr =
   |> Array.indexed
   |> Array.filter (snd >> cond)
   |> function
-    | [| index, _item |] -> index
-    | _ -> failwith "Cannot find required aggregator for executor hook"
+    | [| index, _item |] -> Some index
+    | _ -> None
 
 let private findAggregator aggFn aggFns =
   aggFns
@@ -18,6 +18,9 @@ let private findAggregator aggFn aggFns =
     | AggregateFunction (fn, _) -> fn = aggFn
     | _ -> false
   )
+  |> function
+    | Some index -> index
+    | None -> failwith "Cannot find required aggregator for executor hook"
 
 let private findSingleNonMatchingColumn max (row1: Row) (row2: Row) =
   let rec compare notEqualAt i =
@@ -27,14 +30,14 @@ let private findSingleNonMatchingColumn max (row1: Row) (row2: Row) =
     else if row1.[i] = row2.[i] then
       // Columns equal, try next one
       compare notEqualAt (i + 1)
-    else if notEqualAt > -1 then
+    else if Option.isSome notEqualAt then
       // A non-equal column already exists, abort
-      -1
+      None
     else
       // Mark current index as not equal and keep going to verify no others are
-      compare i (i + 1)
+      compare (Some i) (i + 1)
 
-  compare -1 0
+  compare None 0
 
 type BucketState =
   {
@@ -45,6 +48,11 @@ type BucketState =
     OutputRow: Row
     mutable IsLowCount: bool
   }
+
+type SiblingPerColumn =
+  | NoBuckets
+  | SingleBucket of BucketState
+  | MultipleBuckets
 
 let private mergeBucket groupingLabelsLength lowCountIndex countIndex destinationBucket fromBucket =
   let lowCountAggregator = destinationBucket.Aggregators.[lowCountIndex]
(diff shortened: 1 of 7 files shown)


edongashi:
We should discuss whether we move this to reference or keep it here.

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -0,0 +1,176 @@
...
+open System
+open OpenDiffix.Core
+
+let private findSingleIndex cond arr =
+  arr
+  |> Array.indexed
+  |> Array.filter (snd >> cond)
+  |> function
+    | [| index, _item |] -> index
+    | _ -> failwith "Cannot find required aggregator for executor hook"
```
cristianberneanu, line 12: The function's name is pretty generic, but the error message is very specific.

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -0,0 +1,176 @@
...
+  let rec compare notEqualAt i =
+    if i = max then
+      // End of comparison, return single index (may be -1)
+      notEqualAt
+    else if row1.[i] = row2.[i] then
+      // Columns equal, try next one
+      compare notEqualAt (i + 1)
+    else if notEqualAt > -1 then
+      // A non-equal column already exists, abort
+      -1
```
cristianberneanu, line 32: How about using `None` instead of `-1`?

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -0,0 +1,176 @@
...
+      OutputRow = Array.init rowSize (fun i -> if i < groupingLabelsLength then group.[i] else Null)
+      IsLowCount = false
+    }
+
+  let state = Collections.Generic.Dictionary<Row, BucketState>(Row.equalityComparer)
+
+  // Group rows into buckets and keep track of low-count/count arguments
+  for row in Executor.execute executionContext childPlan do
+    let group = groupingLabels |> Array.map (Expression.evaluate row)
+    let argEvaluator = Expression.evaluate row
```
cristianberneanu, line 96: Inconsistent usage: either use this in the previous line as well, or expand the call everywhere.

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -0,0 +1,176 @@
...
+        let bucketState = makeBucket group
+        state.[group] <- bucketState
+        bucketState
+
+    bucket.Aggregators
+    |> Array.iteri (fun i aggregator ->
+      let args = aggArgs.[i] |> List.map argEvaluator
+
+      if i = lowCountIndex then bucket.LowCountArguments.Add(args)
+      else if i = countIndex then bucket.CountArguments.Add(args)
```
cristianberneanu, line 111: It would be more efficient here (but also more complex overall) to instead provide a generic way in `reference` to merge internal aggregator states.

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -0,0 +1,176 @@
...
+      bucket
+    )
+    |> Seq.toArray
+
+  // Merge victim buckets into sibling buckets
+  buckets
+  |> Array.iteri (fun i victimBucket ->
+    if victimBucket.IsLowCount then
+      let victimRow = victimBucket.OutputRow
+      let siblingsPerColumn = Array.init groupingLabelsLength (fun _i -> Collections.Generic.List<BucketState>(2))
```
cristianberneanu, line 143: Why is it worth using `Collections.Generic.List` here instead of the native `list` or `array`?

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -0,0 +1,176 @@
...
+            let listOfSiblings = siblingsPerColumn.[nonMatchingColumn]
+            // We don't need to store all, just whether there are 0, 1, or more siblings
+            if listOfSiblings.Count < 2 then listOfSiblings.Add(otherBucket)
+      )
+
+      let hasUnknownColumn = siblingsPerColumn |> Array.exists (fun list -> list.Count = 0)
+
+      if hasUnknownColumn then
+        siblingsPerColumn
+        |> Array.iter (fun list ->
```
cristianberneanu, line 160: A `for` loop would work well here.

cristianberneanu (commented):
The LED algorithm is a bit convoluted, I think it would help with figuring out what is happening if you put it in a comment somewhere in the module.

anonymizer/OpenDiffix.Service/Led.fs
edongashi: Duplicated code from reference :confused:
We need a more generic approach like you said.

cristianberneanu:
> We should discuss whether we move this to reference or keep it here.

I don't like that much that part of the aggregation logic is duplicated here, so I would prefer to have this in `reference` in the long run. But we need to first clarify our future plans for LED and see if they are compatible with this version. It makes no sense to move the code only to rewrite it later.

anonymizer/OpenDiffix.Service/Led.fs
edongashi: You mean an `IAggregator` should support a `.Merge` function?

anonymizer/OpenDiffix.Service/Led.fs
edongashi: We need to keep track of how many buckets are matching. This could as well be a discriminated union with 3 cases: `NoBucket | SingleBucket bkt | MultipleBuckets`

anonymizer/OpenDiffix.Service/Led.fs
cristianberneanu: Yes.

anonymizer/OpenDiffix.Service/Led.fs
cristianberneanu: Either that or a regular list/array. I find the usage of `Collections.Generic.List` confusing.

edongashi: I will convert to union because it should be much more readable.

anonymizer/OpenDiffix.Service/Led.fs
cristianberneanu: With that we can afterwards make a generic `merge` function for buckets.
And instead of having a full `Aggregation` hook, we could have a lighter `PostAggregation` one.

edongashi: With the funny internal aggregator state we hold I was hesitant to fiddle there. LowCount appears easy enough, as for count and count distinct we'll have to look into it more carefully.

anonymizer/OpenDiffix.Service/Led.fs
edongashi: With the pattern match `iter` works better now.

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -136,6 +136,19 @@ let private executeLed executionContext (childPlan, groupingLabels, aggregators)
...
+  //
+  // for each low count bucket:
+  //   keep track of siblings per column (initially all empty lists)
+  //
+  //   for each other bucket:
+  //     find index where a single column is different
+  //     if such index:
+  //       add bucket to that column's siblings
+  //
+  //   if column siblings contains an empty list:
```
cristianberneanu, line 156: maybe "if a column without siblings exists"?

cristianberneanu (approved):

cristianberneanu (approved):

PR #260, opened 2021-11-17
Add support for global aggregation.

### LINKED ISSUE #221: Add support for global aggregation.
When no column is selected for anonymization, we could show the anonymization result of global aggregation.

cristianberneanu:
Closes #221.

### DIFF
--- a/anonymizer/OpenDiffix.Service/Program.fs
+++ b/anonymizer/OpenDiffix.Service/Program.fs
@@ -64,15 +64,22 @@ let handlePreview
     | Rows -> "*"
     | Entities -> $"distinct %s{aidColumn}"
 
-  let query =
+  let countColumns =
     $"""
-      SELECT
-        diffix_low_count(%s{aidColumn}),
-        count(%s{countInput}), diffix_count(%s{countInput}, %s{aidColumn}),
-        %s{String.join ", " buckets}
-      FROM table
-      GROUP BY %s{String.join ", " [ 4 .. buckets.Length + 3 ]}
+      diffix_low_count(%s{aidColumn}),
+      count(%s{countInput}),
+      diffix_count(%s{countInput}, %s{aidColumn})
     """
+
+  let query =
+    if Array.isEmpty buckets then
+      $"SELECT %s{countColumns} FROM table"
+    else
+      $"""
+        SELECT %s{countColumns}, %s{String.join ", " buckets}
+        FROM table
+        GROUP BY %s{String.join ", " [ 4 .. buckets.Length + 3 ]}
+      """
 
   let result = runQuery query inputPath anonParams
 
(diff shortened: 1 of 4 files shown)


anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -134,15 +141,18 @@ let handleExport
     | Rows -> "*"
     | Entities -> $"distinct %s{aidColumn}"
 
+  let countColumn = $"diffix_count(%s{countInput}, %s{aidColumn}) AS count"
```
pdobacz, line 144: `countColumns` too?

pdobacz (approved):
LGTM + small nitpick

anonymizer/OpenDiffix.Service/Program.fs
cristianberneanu: I think I prefer the singular: there is only one column here and we don't have plans to add others.

pdobacz: I'm blind :see_no_evil:

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -242,15 +242,9 @@ export const ColumnSelectionStep: FunctionComponent<ColumnSelectionStepProps> =
...
-        {bucketColumns.length !== 0 && (
-          <>
-            <Divider />
-            {children({ bucketColumns, countInput: hasAidColumn ? countInput : 'Rows' })}
-          </>
-        )}
-      </div>
+      {/* Render next step */}
+      <Divider />
+      {children({ bucketColumns, countInput: hasAidColumn ? countInput : 'Rows' })}
```
edongashi, line 247: Shame that we overwhelm the user with 4 blue nav steps as soon as they load a file. Maybe we should re-think the "steps" idea.

pdobacz: Hm, and what is the intended use of the global aggregation? Is it
1/ actually allow to release **anonymized** counts of either rows or entities
2/ just give an overview of what kind of data is handled?

If (2/) then maybe these two values should be presented on the `Data Preview` step, without anonymization?

cristianberneanu: Yes, not ideal. The alternative would be to make the user press "Next" after each step. This would make first runs less overwhelming, but it is probably pretty annoying in the long run.
Let's postpone this discussion until after the first full release.

cristianberneanu: > Hm, and what is the intended use of the global aggregation?

I think it provides an overview over the input data and over the effect anonymization has on it.

pdobacz: > > Hm, and what is the intended use of the global aggregation?
> 
> I think it provides an overview of the input data and of the effect anonymization has on it.

Yeah, my point being here - is the anonymization of the entire table interesting that often, except pathological cases where the table has very few rows/entities?

If this effect is sometimes, but rarely interesting, we could have non-anonymized counts as part of the `Data Preview` and then if anyone wants the global anonymized aggregates, there could be a special `None` switch in the `Column Selection` step, which would need to be explicitly switched on to give the user the global anonymized aggregates.

cristianberneanu: I find it interesting enough and Paul also seems to agree (he asked for it).
I don't like branches or special cases in the usage flow. The current approach has the big advantage that is consistent: the anonymization step runs the same regardless if you select something or not in the previous step.

PR #261, opened 2021-11-17
Allow passing custom anonParams to the OpenDiffix.Service

pdobacz:
Behavior of Diffix for Desktop remains unchanged - since `anonParams` can be omitted in the `Preview` and `Export` requests, everything should work the same on the GUI.

The next step is to supply `anonParams` from the GUI and (possibly) turn off the defaults used on `OpenDiffix.Service` end.

One tiny spot I don't necessarily like is that our types `CommonTypes.AnonymizationParams` and `RequestAnonParams` diverge slightly. I think this is still acceptable - we don't necessarily want to allow custom `TableSettings` to slip in, which would be possible if we used the format of `CommonTypes.AnonymizationParams` for the JSON communication. The way `salt` is calculated in desktop is also slightly separate from the remainder of the params, so I've decided to keep it separate in the JSON API.

I'm taking the liberty to roll with this PR as it is now, postponing the discussion about where and how exactly between `desktop`, `service` and `reference` will the default `anonParams` "live"

### DIFF
--- a/anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
+++ b/anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
@@ -15,7 +15,9 @@ type TempFile() =
     member this.Dispose() = System.IO.File.Delete(path)
 
 let defaultAnonParams =
-  $"""{{"suppression":{{"lowThreshold":3,"sD":1,"lowMeanGap":2}},"outlierCount":{{"lower":2,"upper":5}},"topCount":{{"lower":2,"upper":5}},"noiseSD":1.0}}"""
+  $"""
+  {{"suppression":{{"lowThreshold":3,"sD":1,"lowMeanGap":2}},"outlierCount":{{"lower":2,"upper":5}},"topCount":{{"lower":2,"upper":5}},"noiseSD":1.0}}
+  """
 
 [<Fact>]
 let ``Handles Load request`` () =
@@ -56,9 +58,14 @@ let ``Handles Preview request`` () =
 
 [<Fact>]
 let ``Handles Preview request with custom anonParams`` () =
+  let anonParams =
+    $"""
+    {{"suppression":{{"lowThreshold":0,"sD":0,"lowMeanGap":0}},"outlierCount":{{"lower":2,"upper":5}},"topCount":{{"lower":2,"upper":5}},"noiseSD":1.0}}
+    """
+
   let requestCustom =
     $"""
-    {{"type":"Preview","inputPath":"%s{dataPath}","aidColumn":"id","salt":"1","anonParams":{{"suppression":{{"lowThreshold":0,"sD":0,"lowMeanGap":0}},"outlierCount":{{"lower":2,"upper":5}},"topCount":{{"lower":2,"upper":5}},"noiseSD":1.0}},"buckets":["age","city"],"countInput":"Rows","rows":1000}}
+    {{"type":"Preview","inputPath":"%s{dataPath}","aidColumn":"id","salt":"1","anonParams":%s{anonParams},"buckets":["age","city"],"countInput":"Rows","rows":1000}}
     """
 
   let responseCustom = requestCustom |> mainCore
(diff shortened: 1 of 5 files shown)


cristianberneanu:
> I'm taking the liberty to roll with this PR as it is now, postponing the discussion about where and how exactly between desktop, service and reference will the default anonParams "live"

The anonymization params should live in `desktop` and they should not be optional in corresponding requests.

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
```diff
@@ -65,3 +91,20 @@ let ``Handles Export request`` () =
...
+
+  let requestCustom =
+    $"""
+    {{"type":"Export","inputPath":"%s{dataPath}","aidColumn":"id","salt":"1","anonParams":{{"suppression":{{"lowThreshold":300,"sD":1,"lowMeanGap":2}},"outlierCount":{{"lower":2,"upper":5}},"topCount":{{"lower":2,"upper":5}},"noiseSD":1.0}},"buckets":["age","city"],"countInput":"Rows","outputPath":"%s{outputFile.Path}"}}
+    """
+
+  requestCustom |> mainCore |> should equal ""
+
+  let result = System.IO.File.ReadAllLines(outputFile.Path)
+  // The custom anonymization params have ridiculous supporesion thesshold to suppress all buckets
```
edongashi, line 107: > supporesion thesshold

Typos

anonymizer/OpenDiffix.Service/JsonEncodersDecoders.fs
```diff
@@ -23,6 +23,16 @@ type CountInput =
   | Rows
   | Entities
 
+type RequestAnonParams =
+  {
+    Suppression: CommonTypes.SuppressionParams
```
edongashi, line 28: Why the `CommonTypes` prefix? The module is auto opened. Unless there is another type with such name?

anonymizer/OpenDiffix.Service/JsonEncodersDecoders.fs
```diff
@@ -31,6 +41,7 @@ type PreviewRequest =
     Salt: string
     Buckets: string []
     CountInput: CountInput
+    AnonParams: Option<RequestAnonParams>
```
edongashi, line 44: Like Cristian noted, we should make these required so that we don't end up with all the loose defaults around.

src/shared/anonymizer.ts
```diff
@@ -111,13 +111,26 @@ class DiffixAnonymizer implements Anonymizer {
         inputPath: schema.file.path,
         aidColumn: `"${aidColumn}"`,
         salt: schema.salt,
+        // TODO: send optional `anonParams` from the UI, template:
```
edongashi, line 114: I'd hardcode this object right now and remove the Option from service.

src/types/index.ts
```diff
@@ -111,6 +111,30 @@ export type AnonymizedAggregate = {
...
+// AnonymizationParams
+
+export type Interval = {
+  lower: number;
+  upper: number;
+};
+
+export type SuppressionParams = {
+  lowThreshold: number;
+  sD: number;
```
edongashi, line 123: `sD` is a weird name, just `sd` or `SD`?

edongashi (dismissed):
OK by me if you plan on gradually migrating to a single source of truth in `desktop` (electron part). Currently it's a bit hard to track what is being assigned where.

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
pdobacz: :see_no_evil:

anonymizer/OpenDiffix.Service/JsonEncodersDecoders.fs
pdobacz: ah leftovers from time where I introduced a `SuppressionParams` here

src/types/index.ts
pdobacz: it's SD in F# world, but the auto JSON decoder with `CamelCase` strategy interprets `sD`. I didn't want to uproot the naming conventions _yet_ (e.g. expand this to `StdDev`) or introduce any special cases in the JSON codecs.

anonymizer/OpenDiffix.Service/JsonEncodersDecoders.fs
pdobacz: ok, I'll handle this in this PR, since we know how to organize these already and I need to correct style anyway.

edongashi (approved):

anonymizer/OpenDiffix.Service.Tests/Program.Tests.fs
```diff
@@ -51,17 +54,46 @@ let ``Handles Preview request`` () =
   response |> should haveSubstring "medianDistortion"
   response |> should haveSubstring "rows"
 
+[<Fact>]
+let ``Handles Preview request with custom anonParams`` () =
+  let requestCustom =
+    $"""
+    {{"type":"Preview","inputPath":"%s{dataPath}","aidColumn":"id","salt":"1","anonParams":{{"suppression":{{"lowThreshold":0,"sD":0,"lowMeanGap":0}},"outlierCount":{{"lower":2,"upper":5}},"topCount":{{"lower":2,"upper":5}},"noiseSD":1.0}},"buckets":["age","city"],"countInput":"Rows","rows":1000}}
```
cristianberneanu, line 61: These lines are pretty long and hard to read. Can you please split them into multiple lines?

edongashi: Also why not send an object to the test? We can assume (or test somewhere else) that JSON parsing works as intended.

pdobacz: I've split the longest ones somewhat, following the structure of the request payload.

For now this is a simple smoke test to ensure the stack works, in particular, that the requests from `anonymizer.ts` work. It's very convenient to copy over the request string from logs into the test. I'm not sure we're investing time into more fine-grained testing at this point.

edongashi (approved):

PR #262, opened 2021-11-18
GUI step for choosing anonymization params

### LINKED ISSUE #249: Support configuration of anonymization parameters
We need to first decide which parameters makes sense to expose and where to place them.
Then we need to update the service request structures.

pdobacz:
Closes #249 

I have one quick thought (which I hinted upon in our earlier discussions) - with current way the results (i.e. Anonymization summary) is presented, it's quite hard to see that changing the params actually makes a change (one has to memorize the 4 values of the summary over the course of recalculating the anonymization). I'd say that making the animation which indicates calculation in progress continue to show previously calculated values or sth.


src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -165,7 +165,7 @@ export const ColumnSelectionStep: FunctionComponent<ColumnSelectionStepProps> =
   return (
     <>
       <div className="ColumnSelectionStep notebook-step">
-        <NotebookNavAnchor step={NotebookNavStep.ColumnSelection} status={anySelected ? 'done' : 'active'} />
+        <NotebookNavAnchor step={NotebookNavStep.ColumnSelection} status='done'/>
```
pdobacz, line 168: I noticed that now, when the ColumnSelection step has a "default" setting (which displays the global aggregation), this must always show as `done`, otherwise we end up with "gaps" in the visual representation of the progress through the steps.

It was hardly noticeable before the AnonParams step, but with it, it became hard to miss.

pdobacz:
> > I'd say that making the animation which indicates calculation in progress continue to show previously calculated values or sth.
> 
> Sounds good. We can show a spinner somewhere else to indicate that they are being recomputed.

I just accidentally noticed that the results table does it like that (spinner without hiding the previous result). I'll use this as a template, so it should be easy enough to include in this pr

edongashi:
The CI build is failing. You can force push if you want to because I haven't started yet.

pdobacz:
> The CI build is failing. You can force push if you want to because I haven't started yet.

yeah, I missed something, sorry about this. Is good to review now, I'll make a separate PR for the spinner thing, as you suggested.

anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -54,7 +54,7 @@ let getAnonParams (requestAnonParams: RequestAnonParams) (salt: string) =
     Suppression = requestAnonParams.Suppression
     OutlierCount = requestAnonParams.OutlierCount
     TopCount = requestAnonParams.TopCount
-    NoiseSD = requestAnonParams.NoiseSD
+    NoiseSD = sqrt requestAnonParams.BaseNoiseSD
```
edongashi, line 57: Why is it passed as squared? It's technically a variance?

docs/operation.md
```diff
@@ -95,6 +95,29 @@ When a column is selected, the generalization input is exposed. For text columns
...
+__Diffix for Desktop__ has 3 presets of anonymization parameters:
+
+| Parameter | Max utility | Balanced | Max privacy |
+| --- | --- | --- | --- |
+| `low_thresh` - The lower bound for the noisy suppression threshold | 3 | 4 | 5 |
+| `sd_supp` - The standard deviation of the suppression Gaussian noise | 1.0 | 1.0 | 1.0 |
+| `low_mean_gap` - The number of standard deviations (`sd_supp`) between `low_thresh` and the Gaussian noise mean | 2 | 2 | 2 |
+| `base_sd` - The standard deviation of the noise| 1.5 | 2.25 | 3.0 |
+| `outlier_min` - The minimum possible value of `outlier_count` | 1 | 2 | 2 |
+| `outlier_max` - The maximum possible  values of `outlier_count` | 2 | 3 | 3 |
```
edongashi, line 115: Double space. `values of` -> `value of` to match with sibling param above.

src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { FunctionComponent, useState } from 'react';
+import { Divider, Radio, Typography, Tooltip } from 'antd';
+
+import { NotebookNavAnchor, NotebookNavStep } from '../Notebook';
+import { AnonParamsPreset, AnonymizationParams } from '../types';
+
+import './AnonParamsSelectionStep.css';
```
edongashi, line 7: Empty CSS can be removed

src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <>
+      <div className="AnonParamsSelectionStep notebook-step">
+        <NotebookNavAnchor step={NotebookNavStep.AnonParamsSelection} status="done" />
+        <Title level={3}>Adjust anonymization parameters</Title>
+        <Radio.Group value={preset} size="large" buttonStyle="solid" onChange={(e) => setPreset(e.target.value)}>
+          <Tooltip placement="bottom" title="Strong anonymity, best output utility">
+            <Radio.Button value="P">Max utility</Radio.Button>
+          </Tooltip>
+          <Tooltip placement="bottom" title="Better anonymity, great output utility">
```
edongashi, line 73: "Better" anonymity can sound like the previous one is worse. Maybe "stronger"?

src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -165,7 +165,7 @@ export const ColumnSelectionStep: FunctionComponent<ColumnSelectionStepProps> =
   return (
     <>
       <div className="ColumnSelectionStep notebook-step">
-        <NotebookNavAnchor step={NotebookNavStep.ColumnSelection} status={anySelected ? 'done' : 'active'} />
+        <NotebookNavAnchor step={NotebookNavStep.ColumnSelection} status="done" />
```
edongashi, line 168: :+1:

src/Notebook/notebook-help.tsx
```diff
@@ -74,6 +74,22 @@ function ColumnSelectionHelp() {
   );
 }
 
+function AnonParamsSelectionHelp() {
+  return (
+    <div>
+      <Title level={4}>Anonymization Parameters</Title>
+      <Paragraph>
+        The default set of parameters ensures correct anonymization and has maximum analytical utility. It is
+        recommended for the majority of use cases. Select stronger anonymization, if it does not lower the analytical
```
edongashi, line 83: "Select stronger anonymization, if it does not lower the analytical..." I feel like the comma makes this like a suggestion instead of a condition.

src/shared/anonymizer.ts
```diff
@@ -104,24 +105,15 @@ class DiffixAnonymizer implements Anonymizer {
...
-          suppression: {
-            lowThreshold: 3,
-            sD: 1.0,
-            lowMeanGap: 2.0,
-          },
-          outlierCount: { lower: 2, upper: 5 },
-          topCount: { lower: 2, upper: 5 },
-          noiseSD: 1.0,
-        },
+        anonParams: anonParams,
```
edongashi, line 116: `anonParams: anonParams` can be shortened to just `anonParams`.

anonymizer/OpenDiffix.Service/Program.fs
pdobacz: embarrassing, should be `BaseNoiseSD / sqrt(2)`, I don't know why I did this, good catch!

src/Notebook/notebook-help.tsx
pdobacz: thanks, without comma is actually the only correct punctuation in English, I always get it wrong :shrug:

anonymizer/OpenDiffix.Service/Program.fs
cristianberneanu: DfD uses total noise, while `reference` uses per-layer noise. Total noise scales with the square root of the number of noise samples. Currently, we have 2 noise layers. Meaning that `TotalSD = LayerSD * sqrt(2)` or `LayerSD = TotalSD / sqrt(2)`.

pdobacz: ah, the reason is that `NoiseSD` is actually `noise_sd_layer`, which is a single layer SD. `base_sd` is for the sum of two noise layers, so the std dev is "split" between these layers.

edongashi (approved):

PR #269, opened 2021-11-24
Make navbar collapsible using our custom Layout component

### LINKED ISSUE #145: Make side nav collapsible

pdobacz:
Closes #145 

This is still looking far from perfect, but I'd like to stop here and evaluate. This is using the current "custom" `Layout` component, which is also used for the Docs panel - it turned out to be less cumbersome to style than the `antd`'s `Layout`.

I also did my best to perform the css voodoo right, but might still missed a ton.

I'm also happy to stop here and give up on the collapsible nav bar, if we deem this is too messy or too rough, more than justified by the benefits of the feature.

Screenshot from 2021-11-24 16-57-54
Screenshot from 2021-11-24 16-57-41

### DIFF
--- a/src/Notebook/Notebook.css
+++ b/src/Notebook/Notebook.css
@@ -3,6 +3,6 @@
 }
 
 .Notebook-content {
-  padding: 0 32px 96px 32px;
+  padding: 0 32px 96px 8px;
   overflow-y: scroll;
 }
--- a/src/Notebook/Notebook.css
+++ b/src/Notebook/Notebook.css
@@ -6,3 +6,20 @@
   padding: 0 32px 96px 32px;
   overflow-y: scroll;
 }
+
+.Notebook-sidebar {
+  display: flex;
+}
+
+.Notebook-sidebar-right,
+.Notebook-sidebar-left {
+  display: inline-block;
+}
+
+.Notebook-sidebar .ant-btn {
+  height: 0;
+  width: 0;
+  padding: 0 5px 0 0;
+  display: inline-block;
+  border: none;
+}
--- a/src/Notebook/Notebook.css
+++ b/src/Notebook/Notebook.css
@@ -6,20 +6,3 @@
   padding: 0 32px 96px 32px;
   overflow-y: scroll;
 }
-
-.Notebook-sidebar {
-  display: flex;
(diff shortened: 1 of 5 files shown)


edongashi:
When it's collapsed it would be better to fix the button outside of the sidebar. Currently it adds this awkward column taking up space and not looking right. For that you'd need ugly CSS hacks or sharing data with context. I think it's not worth adding that complexity. Let's put this on hold?

pdobacz:
> Currently it adds this awkward column taking up space and not looking right. For that you'd need ugly CSS hacks or sharing data with context. I think it's not worth adding that complexity. Let's put this on hold?

yeah, I agree, the extra button kinda defies the point of collapsing. I tried also putting it above the nav bar contents, but it didn't look good.

One late idea I had, inspired by slack, is to have the nav bar collapse, when the user "grabs" the border and moves to the left (the channel sidebar in slack does this).

I can do some more attempts in between other stuffs, actually I've got a bunch of ideas now. Thank you for the comment!

edongashi:
[This example](https://ant.design/components/layout/#components-layout-demo-custom-trigger)  has an external trigger which I like.

pdobacz:
This turned out to be a quick win:
Screenshot from 2021-11-25 10-05-31
Screenshot from 2021-11-25 10-05-27

edongashi:
> This turned out to be a quick win:

With such a wide navbar the button looks unrelated 😕 . Maybe the [next example](https://ant.design/components/layout/#components-layout-demo-responsive) where it's fixed to the sidebar?

pdobacz:
Screenshot from 2021-11-25 11-09-08

:sweat_smile:

edongashi:
In a box and with a shadow it creates depth

image

pdobacz:
Screenshot from 2021-11-25 11-45-43

OK, this works too for me (I just don't like that the button obscures the contents as in the example, I made it not do that.

pdobacz:
both Edon and I are concerned about:
 - the complexity this brings vs low benefit
 - the obscurity of the icons (caused by the non-standard steps we have)

Erring on the side of simplicity, let's not have this and close until there's more arguments for having this.

PR #275, opened 2021-11-30
Make only Suppression Threshold configurable

### LINKED ISSUE #273: Modification to Desktop-settable anonymization parameters
Having completed the analysis for the Elm doc, I'm now of the strong opinion that the only really important anonymization parameter we need to expose in Desktop is `low_thresh`.

So instead of the `P`, `XP`, `XXP` family of settings, I propose a setting which we can call "Suppression Threshold" (which internally maps to `low_thresh`).

The default would be 3 (meaning that buckets with 2 or 1 AIDV are suppressed, but 3 may not be suppressed).

I would like the minimum to be 2. I don't see a need for a maximum (though in practice 10 or higher would be pretty excessive).

Maybe this kind of thing:
image

pdobacz:
Closes #273

### DIFF
--- a/docs/anonymization.md
+++ b/docs/anonymization.md
@@ -38,7 +38,7 @@ __Diffix__ recognizes how many protected entities contribute to each output bin.
 
 Rather than apply a single suppression threshold to all bins, __Diffix__ slightly modifies the threshold for different bins. This adds additional uncertainty for recipients that have prior knowledge of protected entities.
 
-By default, __Diffix__ suppresses bins with fewer than 5 protected entities *on average*. This corresponds to the `Suppression Threshold` anonymization parameter equal 3, with the average modification added. This can be configured (see section [Configure anonymization parameters](operation.md#configure-anonymization-parameters)). A minimum value 2 of `Suppression Threshold` is enforced, therefore __Diffix__ always suppresses bins with only a single contributing protected entity.
+By default, __Diffix__ suppresses bins with fewer than 5 protected entities *on average*, corresponding to the `Suppression Threshold` anonymization parameter default value of 3. This can be configured at runtime (see section [Configure anonymization parameters](operation.md#configure-anonymization-parameters)). A minimum value of 2 for `Suppression Threshold` is enforced, therefore __Diffix__ always suppresses bins with only a single contributing protected entity.
 
 ## Generalization
 
--- a/docs/anonymization.md
+++ b/docs/anonymization.md
@@ -38,7 +38,7 @@ __Diffix__ recognizes how many protected entities contribute to each output bin.
 
 Rather than apply a single suppression threshold to all bins, __Diffix__ slightly modifies the threshold for different bins. This adds additional uncertainty for recipients that have prior knowledge of protected entities.
 
-__Diffix__ suppresses bins with fewer than 4 protected entities *on average*. __Diffix__ always suppresses bins with only a single contributing protected entity.
+By default, __Diffix__ suppresses bins with fewer than 5 protected entities *on average*. This corresponds to the `Suppression Threshold` anonymization parameter equal 3, with the average modification added. This can be configured (see section [Configure anonymization parameters](operation.md#configure-anonymization-parameters)). A minimum value 2 of `Suppression Threshold` is enforced, therefore __Diffix__ always suppresses bins with only a single contributing protected entity.
 
 ## Generalization
 
--- a/docs/operation.md
+++ b/docs/operation.md
@@ -97,24 +97,24 @@ If the input data is multi-row, then __Diffix for Desktop__ gives you the choice
 
 ## Configure anonymization parameters
 
-The default preset ("Max utility") consists of the recommended values for the default anonymization parameters, which have been found to provide strong anonymity and maximize the analytical utility of the resulting data.
+The default preset consists of the recommended values for the anonymization parameters, which have been found to provide strong anonymity and maximize the analytical utility of the resulting data.
 
-If you are forced to increase the strength of anonymization, or find that stronger anonymization still produces satisfactory results in terms of data quality, you can opt to choose one of the stronger presets.
+One of the parameters (`low_thresh` - Suppression Threshold) is made configurable. Decrease it, to allow more of the smallest bins in the anonymized data and improve utility. Increase it, to suppress more of the smallest bins and increase privacy.
 
-### Anonymization parameters presets
+The remainder of parameters are currently fixed at the recommended default values.
 
-__Diffix for Desktop__ has 3 presets of anonymization parameters:
+__Diffix for Desktop__ uses the following anonymization parameters:
 
-| Parameter                                                                                                       | Max utility | Balanced | Max privacy |
-| --------------------------------------------------------------------------------------------------------------- | ----------- | -------- | ----------- |
-| `low_thresh` - The lower bound for the noisy suppression threshold                                              | 3           | 4        | 5           |
-| `sd_supp` - The standard deviation of the suppression Gaussian noise                                            | 1.0         | 1.0      | 1.0         |
-| `low_mean_gap` - The number of standard deviations (`sd_supp`) between `low_thresh` and the Gaussian noise mean | 2           | 2        | 2           |
-| `base_sd` - The standard deviation of the noise                                                                 | 1.5         | 2.25     | 3.0         |
-| `outlier_min` - The minimum possible value of `outlier_count`                                                   | 1           | 2        | 2           |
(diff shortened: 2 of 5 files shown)


docs/operation.md
```diff
@@ -97,24 +97,24 @@ If the input data is multi-row, then __Diffix for Desktop__ gives you the choice
...
+| `low_thresh` - The lower bound for the noisy suppression threshold                                              | 3           | yes           |
+| `sd_supp` - The standard deviation of the suppression Gaussian noise                                            | 1.0         | no            |
+| `low_mean_gap` - The number of standard deviations (`sd_supp`) between `low_thresh` and the Gaussian noise mean | 2           | no            |
+| `base_sd` - The standard deviation of the noise                                                                 | 1.5         | no            |
+| `outlier_min` - The minimum possible value of `outlier_count`                                                   | 1           | no            |
+| `outlier_max` - The maximum possible value of `outlier_count`                                                   | 2           | no            |
+| `top_min` - The minimum possible value of `top_count`                                                           | 3           | no            |
+| `top_max` - The maximum possible value of `top_count`                                                           | 4           | no            |
```
cristianberneanu, line 117: It probably doesn't make sense to mention these parameters anymore. The information is not actionable anyway.

src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
```diff
@@ -14,79 +14,42 @@ export type AnonParamsSelectionStepData = {
...
+  const anonParams = {
+    suppression: {
+      lowThreshold: lowThreshold,
+      sD: 1.0,
+      lowMeanGap: 2.0,
+    },
+    outlierCount: { lower: 1, upper: 2 },
+    topCount: { lower: 3, upper: 4 },
+    baseNoiseSD: 1.5,
+  };
```
cristianberneanu, line 30: Can you also please cleanup the anon params request structure? It serves no purpose to duplicate the default settings here if they can't vary at all.

cristianberneanu (approved):

docs/operation.md
pdobacz: I left it in on purpose, given the previous discussions, that we will provide configurability if needed. In this context I think it's good to inform about the broad choice somewhere. So, the action a user might take is to request configurability .Also this serves as documentation of the values to compare other materials (e.g. paper). 

WDYT?

src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
pdobacz: in similar vein to the comment on docs - this is on purpose. Earlier we've made the decision to keep the defaults which apply for DfD in here, so I think we should follow this decision.

Otherwise we'll end up with the defaults being spread across the code base. The fact that only `low_thresh` is configurable sounds like being purely up to the DfD GUI.

docs/operation.md
cristianberneanu: This should be documentation for the app, not for the Diffix mechanism. And the app doesn't use these parameters.
As things are, I think this hinders, more than it helps.

src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
cristianberneanu: The defaults are spread across the code base now. We have these values in `reference` and also here. So any change to them needs to touch both repos. I would like to avoid that.

> we've made the decision to keep the defaults which apply for DfD in here

I understood things differently: we made a decision to keep the various presets here, one of which was chosen as the default one. The option to keep them in `reference` wasn't available, as the concept of preset settings is present there.

pdobacz: > The defaults are spread across the code base now. We have these values in reference and also here. So any change to them needs to touch both repos. I would like to avoid that.

I meant that all the values which apply for DfD are currently all in DfD and was referring to [this thread](https://github.com/diffix/desktop/pull/261#issuecomment-971757801)

hm, FWIW, I think it's 1/ not worth the legwork 2/ less consistent this way. I'd prefer all-or-nothing - if we set one parameter, let's set all. This way it's really easy to tweak values and configurability and iterate faster. And we can reasonably anticipate change, since we're iterating already.

But if you've got a strong opinion about this, let's change. (also mind that the current default for `low_thresh` in reference is 2 and also `top/outlierCount` are different, so we'd need to adjust there as well). Or maybe we can plant a ticket to revisit this and cleanup some time from now?

cristianberneanu: > I meant that all the values which apply for DfD are currently all in DfD and was referring to this thread

Fair enough. But when I made that comment, all the values could vary, now most of them are fixed. To me, it makes most sense to put the default value next to the control logic (which doesn't exist anymore for most of the settings).

But ok, since you say that the `reference` defaults differ, let's make these the canonical ones.

docs/operation.md
pdobacz: I'm not sure about this (right now the users who look for the anonymization parameters used and referred to in the papers, won't know that some values were used at all), but I went ahead and removed the table.

I also tweaked a section in the other page, which I think was out-of-date. please take a look 47f1423d

src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
pdobacz: ok, leaving as is for now and https://github.com/diffix/desktop/issues/277 for some close future

docs/operation.md
```diff
@@ -99,25 +99,10 @@ If the input data is multi-row, then __Diffix for Desktop__ gives you the choice
 
 The default preset consists of the recommended values for the anonymization parameters, which have been found to provide strong anonymity and maximize the analytical utility of the resulting data.
 
-One of the parameters (`low_thresh` - Suppression Threshold) is made configurable. Decrease it, to allow more of the smallest bins in the anonymized data and improve utility. Increase it, to suppress more of the smallest bins and increase privacy.
+One of the parameters (`low_thresh` - Suppression Threshold) is made configurable. Decrease it, to allow more of the smallest bins in the anonymized data and improve utility. Increase it, to suppress more of the smallest bins and increase privacy. The default values is `3` and minimum value is `2`.
```
cristianberneanu, line 102: "default values" -> "default value"

docs/anonymization.md
```diff
@@ -38,7 +38,7 @@ __Diffix__ recognizes how many protected entities contribute to each output bin.
 
 Rather than apply a single suppression threshold to all bins, __Diffix__ slightly modifies the threshold for different bins. This adds additional uncertainty for recipients that have prior knowledge of protected entities.
 
-__Diffix__ suppresses bins with fewer than 4 protected entities *on average*. __Diffix__ always suppresses bins with only a single contributing protected entity.
+By default, __Diffix__ suppresses bins with fewer than 5 protected entities *on average*. This corresponds to the `Suppression Threshold` anonymization parameter equal 3, with the average modification added. This can be configured (see section [Configure anonymization parameters](operation.md#configure-anonymization-parameters)). A minimum value 2 of `Suppression Threshold` is enforced, therefore __Diffix__ always suppresses bins with only a single contributing protected entity.
```
cristianberneanu, line 41: "By default, __Diffix__ suppresses bins with fewer than 5 protected entities *on average*, corresponding to the `Suppression Threshold` anonymization parameter default value of 3."
"A minimum value of 2 for `Suppression Threshold` is enforced"

cristianberneanu (approved):

docs/operation.md
cristianberneanu: Maybe you're right. Paul will go over this part soon and he can add it back.

docs/anonymization.md
pdobacz: it was super difficult for me to do the wordsmithy here.

Are you saying just these 2 sentences is enough? that would be great.

cristianberneanu: Not exactly, I think it reads better with these small tweaks. So the full section should be:

```
By default, __Diffix__ suppresses bins with fewer than 5 protected entities *on average*, corresponding to the `Suppression Threshold` anonymization parameter default value of 3. This can be configured at runtime (see section [Configure anonymization parameters](operation.md#configure-anonymization-parameters)). A minimum value of 2 for `Suppression Threshold` is enforced, therefore __Diffix__ always suppresses bins with only a single contributing protected entity.
```

cristianberneanu (approved):

PR #280, opened 2021-12-01
Migrate LED hook and add basic star bucket hook

edongashi:
The star bucket is retrieved through a side channel and does not affect query pipeline.
Its values are not sent anywhere. I'm not even sure if it works as expected as I haven't tested it.

There's a couple of TODOs which are good starting points. Also tests are a TODO.

I suggest you review commits individually to be aware what goes with LED and what with *.


anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -99,7 +104,22 @@ let handlePreview
...
       """
 
-  let result = runQuery [ ledHook ] query inputPath anonParams
+  // We get a hold of the star bucket reference via side effects.
+  let mutable starBucket: Bucket option = None
+  let starBucketHook = AggregationHooks.StarBucket.hook (fun bucket -> starBucket <- Some bucket)
+
+  let result = runQuery [ ledHook; starBucketHook ] query inputPath anonParams
+
+  // Should be set once the query completes.
```
pdobacz, line 113: this is just a runtime assertion, in lieu of proper tests, correct?

pdobacz (approved):
ok, I just have some optional requests for the LED part, and the * part I'm taking over, so if I get stuck I'll ask you anyway. I can clean-up whatever I find, while I work on it.

I greatly appreciate you making this "starter pack", it will be much much easier to complete this now that I know how the bricks are put together. :pray:

anonymizer/OpenDiffix.Service/AggregationHooks.fs
```diff
@@ -0,0 +1,127 @@
...
+    let anonAggregators = anonymizingAggregatorIndexes aggregationContext
+
+    let buckets =
+      buckets
+      |> Seq.map (fun bucket -> bucket, isLowCount lowCountIndex bucket)
+      |> Seq.toArray
+
+    buckets
+    |> Array.iteri (fun i (victimBucket, isLowCount) ->
+      if isLowCount then
```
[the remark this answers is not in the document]
edongashi: I plan on rewriting LED so this will be obsolete soon. For now just know that this checks for some very specific conditions which won't happen 99% of the time. If a victim bucket is found under those conditions, it is merged somewhere else to "hide in the crowd".

You should only worry about `module StarBucket` right now because the `Led` code will change and it's not worth knowing the implementation details right now.

anonymizer/OpenDiffix.Service/Program.fs
edongashi: What we have done is a little unorthodox. We use the callback trick to assign a mutable local variable (initially `None`) to our result. When unwrapping we have no choice but to fail in case it is `None`. No tests are needed because if it doesn't run _we will know_ from the crash.

I guess a test can assert the same thing and expect the star bucket to have specific stats.
In the tests file I would be happy if we can refactor in 2 modules similar to here, because the setup logic is becoming complex and knowing what is what helps a lot.


Currently `AggregationHooks.fs` is organized like this:

```
shared helpers

module Led =
  led helpers
  let hook = ....

module StarBucket
  star helpers
  let hook = ...
```

PR #283, opened 2021-12-06
Temporarily disable LED hook

edongashi:


anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -108,7 +108,7 @@ let handlePreview
   let mutable starBucket: Bucket option = None
   let starBucketHook = AggregationHooks.StarBucket.hook (fun bucket -> starBucket <- Some bucket)
 
-  let result = runQuery [ ledHook; starBucketHook ] query inputPath anonParams
+  let result = runQuery [ (* ledHook; *) starBucketHook ] query inputPath anonParams
```
cristianberneanu, line 111: Can you also disable the `starBucketHook`? It doesn't seem stable enough to include it in a release.

cristianberneanu (changes requested):

anonymizer/OpenDiffix.Service/Program.fs
edongashi: Yeah, good idea since we plan an immediate release.

cristianberneanu (approved):

PR #291, opened 2021-12-08
Fix bug and standardize code for numeric inputs

### LINKED ISSUE #279: Suppression threshold is a controlled input but has a defaultValue
`defaultValue` and `value` cannot coexist.

pdobacz:
This PR does two things

1/ Closes #279 

It seems [using `initialValues` and `onValuesChange` on a `Form` component is the standard way](https://ant.design/components/form/#Why-is-component-defaultValue-not-working-when-inside-Form.Item). This gives us correct behavior of the Suppression Threshold input box's value

**NOTE** I tried to conform to this pattern in `ColumnSelectionStep` to no avail - something makes it behave funky, if we use this pattern. But since that in turn doesn't require `defaultValue` like the Suppression Threshold, I think it's best to keep as is

2/ as a bonus it fixes a bug, whereby if you input a decimal (`2.5`) in one of the available fields requiring an integer (Suppression Threshold, Substring start or Substring length), it will result in failed anonymization with a large :red_circle:. The fix causes it to round the inputs from the user and also keep things consistent in the UI. It's not super elegant, but out of number of things I've tried, only this works 100%.

### DIFF
--- a/src/ColumnSelectionStep/ColumnSelectionStep.tsx
+++ b/src/ColumnSelectionStep/ColumnSelectionStep.tsx
@@ -97,21 +97,25 @@ function GeneralizationControls({
           <Form.Item label="Substring start" name="substringStart">
             <span key={column.resetCount}>
               <InputNumber
+                // `precision` and `Math.round` in change handler both needed to protect from decimals
+                precision={0}
                 size="small"
                 placeholder="1"
                 min={1}
                 value={column.substringStart as number}
-                onChange={(substringStart) => updateColumn({ substringStart })}
+                onChange={(substringStart) => updateColumn({ substringStart: Math.round(substringStart) })}
               />
             </span>
           </Form.Item>
           <Form.Item label="Substring length" name="substringLength">
             <span key={column.resetCount}>
               <InputNumber
+                // `precision` and `Math.round` in change handler both needed to protect from decimals
+                precision={0}
                 size="small"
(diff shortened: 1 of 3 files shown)


src/ColumnSelectionStep/ColumnSelectionStep.tsx
```diff
@@ -97,21 +97,25 @@ function GeneralizationControls({
           <Form.Item label="Substring start" name="substringStart">
             <span key={column.resetCount}>
               <InputNumber
+                // `precision` and `Math.round` in change handler both needed to protect from decimals
+                precision={0}
                 size="small"
                 placeholder="1"
                 min={1}
                 value={column.substringStart as number}
```
edongashi, line 105: Why does `value` work here but not in threshold selection? Weird...

edongashi (approved):

edongashi:
Like you said somewhere before, it's due to the generalization inputs starting off empty. I guess the number doesn't like to be told what to have initially.

src/ColumnSelectionStep/ColumnSelectionStep.tsx
pdobacz: I would be guessing that the difference comes from the different context of rendering these components, but I have no idea how so. It seems `Form` is quite messy, I've found lots of threads and issues related to thing not working as expected there.

PR #290, opened 2021-12-08
Integrate DfD GUI with star bucket hook

pdobacz:
Closes #250 

There's a number of things to fix after this, but I'd like to postpone until next PR(s), so that this doesn't bloat infinitely:
- [ ] have more tests for `AggregationHooks` for the StarBucket hook, mainly add a test on the interaction between led and star bucket hooks (merged buckets don't end up in the star bucket)
- [ ] refactor the tests/code of the `AggregationHooks` into separate files + `Utils.fs` (which Edon is working on)
- [ ] implement the extension discussed on slack, whereby we'll include `SuppressedAnonLowCount: true | false | null` field in the summary returned from the `StarBucket` hook and the `OpenDiffix.Service` `Summary` JSON. I decided to postpone this, as this might require a non-trivial change to `reference`. I'll make a separate optional PR/issue for that.
- [ ] also #289 , point 2 and seeding spec are not yet observed

PR #292, opened 2021-12-13
Refactor AnonymizedResultsTable row and column types

pdobacz:
When doing #290 I found it a bit hard to "match" the row data structure with the one required by the column-building function. I was constantly getting `undefined has no toString field` errors, because I was fetching `2_anon` instead of `3_anon` field, for example.

This was mostly caused by the "loose" indexing by column index, which trickled down from the "raw" query result coming from the `OpenDiffix.Service`.

This PR attempts to fix this, by aligning the row types:
1/ raw row comes from the service and is transformed to a "named" structure `AnonymizedResultRow` as soon as possible
2/ that structure is next explicitly translated to a "flat" `TableRowData`, which the table renderer will understand. I've added some comments also, which I would find helpful when doing #290.

### DIFF
--- a/src/AnonymizationStep/AnonymizedResultsTable.tsx
+++ b/src/AnonymizationStep/AnonymizedResultsTable.tsx
@@ -11,18 +11,21 @@ import {
   RowData,
   Value,
   BucketColumn,
-  AnonymizedValue,
 } from '../types';
 import { DisplayModeSwitch } from './DisplayModeSwitch';
 
 import './AnonymizedResultsTable.css';
 
-type TableRowData = RowData & {
-  key: number;
-  lowCount: AnonymizedResultRow['lowCount'];
-};
-
 // Columns
+
+type TableColumnData = {
+  title: string;
+  // must refer to a field in `TableRowData`
+  dataIndex: 'anon' | 'real' | 'diff' | string;
+  render: (v: Value) => React.ReactNode;
+  sorter: (rowA: RowData, rowB: RowData) => number;
+  ellipsis: { showTitle: boolean };
+};
 
 // with `ellipsis: { showTitle: false }` set for the column, we're disabling the default tooltip
 // so that the styling of tooltips is more consistent. This forces us to provide Tooltip in all cases
@@ -86,7 +89,12 @@ function renderRelativeNoiseValue(v: Value) {
   return plainStringCell(v === null ? '-' : formatPercentage(v as number));
 }
 
-function makeColumnData(title: string, dataIndex: string, type: ColumnType, render: (v: Value) => React.ReactNode) {
+function makeColumnData(
+  title: string,
+  dataIndex: string,
+  type: ColumnType,
+  render: (v: Value) => React.ReactNode,
+): TableColumnData {
   return {
     title,
     dataIndex,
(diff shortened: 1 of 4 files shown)


src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -98,26 +106,38 @@ function makeColumnData(title: string, dataIndex: string, type: ColumnType, rend
...
 // Rows
 
+// Represents data shown in the table; contains values which are displayed along
+// with some metadata used to style and render the row.
+type TableRowData = RowData & {
+  // RowData contains fields derived from bucket columns,
+  // which are shown in the table along with:
+  anon: number | null;
+  real: number;
+  diff: number | null;
```
edongashi, line 135: This is a step back, because with this we now only support one single aggregator per output row. The moment we add a different aggregator this will become inconsistent.

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -98,26 +106,38 @@ function makeColumnData(title: string, dataIndex: string, type: ColumnType, rend
...
-  (mode: DisplayMode, bucketColumns: BucketColumn[]) => (column: AnonymizedResultColumn, columnIdx: number) => {
-    if (column.type === 'aggregate') {
-      switch (mode) {
-        case 'anonymized':
-          return [makeColumnData(column.name, columnIdx + '_anon', AGG_COLUMN_TYPE, renderValue)];
-        case 'combined':
-          return [
-            makeColumnData(column.name + ' (anonymized)', columnIdx + '_anon', AGG_COLUMN_TYPE, renderLowCountValue),
-            makeColumnData(column.name + ' (original)', columnIdx + '_real', AGG_COLUMN_TYPE, renderValue),
-            makeColumnData('Distortion', columnIdx + '_diff', 'real', renderRelativeNoiseValue),
```
edongashi, line 111: If we want some type safety here, we could use [template literals](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html):

```typescript
type ValueKey = number
type AggregatorKey = `${number}_anon` | `${number}_real` | `${number}_diff`
type Key = ValueKey | AggregatorKey
```

edongashi: The only hard coded value should be `lowCount` as it has special meaning for our UI. Everything else should be rendered automatically.

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -11,19 +11,22 @@ import {
...
-  AnonymizedValue,
 } from '../types';
 import { DisplayModeSwitch } from './DisplayModeSwitch';
 
 import './AnonymizedResultsTable.css';
 
-type TableRowData = RowData & {
-  key: number;
-  lowCount: AnonymizedResultRow['lowCount'];
-};
```
edongashi, line 23: See [this](https://github.com/diffix/desktop/pull/292/files#r767504539)

src/types/index.ts
```diff
@@ -103,13 +103,13 @@ export type AnonymizedResultColumn = {
...
-  realValue: number;
-  anonValue: number | null;
+// Represents row data returned from the preview query request to the
+// `OpenDiffix.Service`
+export type AnonymizedResultRow = {
+  lowCount: boolean;
+  count: number;
+  diffixCount: number | null;
+  bucketValues: RowData;
 };
```
edongashi, line 113: The star bucket could be returned from `OpenDiffix.Service` to match other rows so that we don't have to build it manually. Then we don't have to care about matching column types and values.

edongashi (changes requested):
I think the better option is to build the star bucket in `service` and put it in the result as a new property. Then the client will just prepend it. Or even the concatenation can be done in service.

pdobacz:
ok, agreed this doesn't work out, closing, possibly in favor of a smaller change

PR #293, opened 2021-12-13
Ensure suppress bin row always comes first in Preview table

pdobacz:
Follow-up to #290. There has been a requirement to have this row always first, regardless of sorting.

### DIFF
--- a/src/AnonymizationStep/AnonymizedResultsTable.tsx
+++ b/src/AnonymizationStep/AnonymizedResultsTable.tsx
@@ -20,9 +20,20 @@ import './AnonymizedResultsTable.css';
 type TableRowData = RowData & {
   key: number;
   lowCount: AnonymizedResultRow['lowCount'];
+  isSuppressBinRow: boolean;
 };
 
 // Columns
+
+// Ensures that suppress bin row always comes first, regardless of sorting.
+// It must be first in the `Table`'s dataSource
+function anonymizedColumnSorter(type: ColumnType, dataIndex: number | string) {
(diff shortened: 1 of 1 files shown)


src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -20,10 +20,21 @@ import './AnonymizedResultsTable.css';
...
 };
 
 // Columns
 
+// Ensures that suppress bin row always comes first, regardless of sorting.
+// It must be first in the `Table`'s dataSource
+function anonymizedColumnSorter(type: ColumnType, dataIndex: number | string) {
+  const commonSorter = columnSorter(type, dataIndex);
+  return (rowA: TableRowData, rowB: TableRowData): number => {
+    if (rowA.isSuppressBinRow || rowB.isSuppressBinRow) return 0;
```
edongashi, line 33: Does 0 work for both directions? Doesn't it make more sense to return `-1` if `rowA is star` or `+1` if `rowB is star`?

edongashi (approved):

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: `-1` / `+1` doesn't work, since this make sthe star row "smallest" / "largest", which is not what we want (then the star bucket's position depends on the order of sorting).

`0` means "keep unsorted order", and the unsorted order is fixed when building the `dataSource` input to the `Table` - there we put the suppress bin row first.

PR #294, opened 2021-12-13
Make AnonymizedResultTable data indices slightly more typesafe

pdobacz:
Supersedes #292. This doesn't fix the discrepancy of the numeric indexing of columns between the returned rows from `Service` and the displayed table, but should make it easier to associate one with the other (and harder to get it wrong).

Kudos to Edon for suggesting the type-template

### DIFF
--- a/src/AnonymizationStep/AnonymizedResultsTable.tsx
+++ b/src/AnonymizationStep/AnonymizedResultsTable.tsx
@@ -12,6 +12,7 @@ import {
   Value,
   BucketColumn,
   AnonymizedValue,
+  RowDataIndex,
 } from '../types';
 import { DisplayModeSwitch } from './DisplayModeSwitch';
 
@@ -27,7 +28,7 @@ type TableRowData = RowData & {
 
 // Ensures that suppress bin row always comes first, regardless of sorting.
 // It must be first in the `Table`'s dataSource
-function anonymizedColumnSorter(type: ColumnType, dataIndex: number | string) {
+function anonymizedColumnSorter(type: ColumnType, dataIndex: RowDataIndex) {
   const commonSorter = columnSorter(type, dataIndex);
   return (rowA: TableRowData, rowB: TableRowData): number => {
     if (rowA.isSuppressBinRow || rowB.isSuppressBinRow) return 0;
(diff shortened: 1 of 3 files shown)


src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -114,17 +120,17 @@ const mapColumn =
...
-          return [makeColumnData(column.name, columnIdx + '_anon', AGG_COLUMN_TYPE, renderValue)];
+          return [makeColumnData(column.name, `${columnIdx}_anon`, AGG_COLUMN_TYPE, renderValue)];
         case 'combined':
           return [
-            makeColumnData(column.name + ' (anonymized)', columnIdx + '_anon', AGG_COLUMN_TYPE, renderLowCountValue),
-            makeColumnData(column.name + ' (original)', columnIdx + '_real', AGG_COLUMN_TYPE, renderValue),
-            makeColumnData('Distortion', columnIdx + '_diff', 'real', renderRelativeNoiseValue),
+            makeColumnData(column.name + ' (anonymized)', `${columnIdx}_anon`, AGG_COLUMN_TYPE, renderLowCountValue),
+            makeColumnData(column.name + ' (original)', `${columnIdx}_real`, AGG_COLUMN_TYPE, renderValue),
+            makeColumnData('Distortion', `${columnIdx}_diff`, AGG_COLUMN_TYPE, renderRelativeNoiseValue),
```
edongashi, line 128: Distortion is always real. But why is count a real I don't know... Either way doesn't affect anything.

src/types/index.ts
```diff
@@ -10,8 +10,11 @@ export type ComputedData<T> =
 
 export type DisplayMode = 'anonymized' | 'combined';
 
+export type AggregateRowDataIndex = `${number}_anon` | `${number}_real` | `${number}_diff`;
+export type RowDataIndex = number | AggregateRowDataIndex;
+
 export type RowData = {
-  [dataIndex in number | string]: Value;
+  [dataIndex in RowDataIndex]: Value;
```
edongashi, line 17: Can this be changed to `[dataIndex: RowDataIndex]: Value`?

edongashi (dismissed):

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: dunno, I just made this consistent :shrug:

edongashi (approved):

src/AnonymizationStep/AnonymizedResultsTable.tsx
edongashi: I'd keep this `real` because it's not an aggregated column, it's statistical information. As for aggs I suppose real is a superset of integer (in JS land there's no int).

PR #295, opened 2021-12-13
Conform to suppression bin specs

### LINKED ISSUE #289: Specification of suppression ("star") bucket
Only make a suppression bucket if:

1. The suppression bucket itself passes the suppression criteria (LCF)
2. The suppression bucket is comprised of at least two suppressed buckets

Seed the suppression bucket as normal with any bucket, but use the symbol `*` as the column value for every column.

pdobacz:
Closes #289 

Starting as draft b/c the `reference` update is required, but not yet merged into master.

This PR is for two independent changes, see commits.

Also, I'm abandoning the idea to "expand" the `OpenDiffix.Service` Summary field by `SuppressedAnonLowCount: boolean` or any other changes like this. The reason is the following - we have added a next rule about the suppress bin (we don't show it if it is built out of just one suppressed bucket). Now, these rules are complex enough, that we can say that it is `Service`, which should only be responsible for releasing (or not) of the *-bucket. It will only serve `SuppressedAnonCount: int option`, and if it's `None`, then it means that _for some reason_, the *-bucket may not be published. This decision is abstracted out of the clients (CSV export and DfD GUI).

### DIFF
--- a/anonymizer/OpenDiffix.Service/AggregationHooks.fs
+++ b/anonymizer/OpenDiffix.Service/AggregationHooks.fs
@@ -164,33 +164,24 @@ module StarBucket =
     starBucket |> Bucket.putAttribute BucketAttributes.IS_STAR_BUCKET (Boolean true)
     starBucket
 
-  /// Iterates and materializes the bucket sequence.
-  let private iterateBuckets f (buckets: Bucket seq) =
-    buckets
-    |> Seq.map (fun bucket ->
-      f bucket
-      bucket
-    )
-    // We do this because re-running a sequence is potentially dangerous if it has side effects.
-    |> Seq.toArray
-
   let private computeStarBucket callback aggregationContext (buckets: Bucket seq) =
     let starBucket = makeStarBucket aggregationContext
     let lowCountIndex = lowCountIndex aggregationContext
     let diffixCountIndex = diffixCountIndex aggregationContext
 
-    let starBucketBuckets =
+    let isInStarBucket bucket =
+      let isAlreadyMerged =
+        bucket
+        |> Bucket.getAttribute BucketAttributes.IS_LED_MERGED
+        |> Value.unwrapBoolean
+
+      not isAlreadyMerged && isLowCount lowCountIndex bucket
+
+    let bucketsInStarBucket =
       buckets
-      |> Seq.filter (fun bucket ->
-        let isAlreadyMerged =
-          bucket
-          |> Bucket.getAttribute BucketAttributes.IS_LED_MERGED
-          |> Value.unwrapBoolean
-
-        not isAlreadyMerged && isLowCount lowCountIndex bucket
-      )
-
-    starBucketBuckets |> Seq.iter (mergeAllAggregatorsInto starBucket)
+      |> Seq.filter isInStarBucket
+      |> Seq.map (mergeAllAggregatorsInto starBucket)
+      |> Seq.length
 
     let executionContext = starBucket.ExecutionContext
 
@@ -202,7 +193,7 @@ module StarBucket =
       // NOTE: we can have a suppress bin consisting of a single suppressed bucket,
       // which won't be suppressed by itself (different noise seed). In such case,
       // we must enforce the suppression manually.
-      if isStarBucketLowCount || (Seq.length starBucketBuckets) = 1 then
+      if isStarBucketLowCount || bucketsInStarBucket < 2 then
         Null
       else
         starBucket.Aggregators.[diffixCountIndex].Final(executionContext)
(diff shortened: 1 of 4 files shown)


anonymizer/OpenDiffix.Service/AggregationHooks.fs
```diff
@@ -148,11 +148,10 @@ module StarBucket =
   let private makeStarBucket (aggregationContext: AggregationContext) =
     let isGlobal = isGlobalAggregation aggregationContext
 
-    // TODO: Do we need a specific noise seed for the star bucket?
     let executionContext = aggregationContext.ExecutionContext
```
edongashi, line 151: I would get rid of this local binding because there's a chance we might accidentally use `aggregationContext.ExecutionContext` instead of `bucket.ExecutionContext`.

edongashi: Ah, I see that this is a short function. I thought it spans all the way below.

anonymizer/OpenDiffix.Service/AggregationHooks.fs
```diff
@@ -189,20 +190,23 @@ module StarBucket =
...
       )
 
     let executionContext = starBucket.ExecutionContext
 
-    // NOTE this is also true when there is no star bucket (no suppression)
-    //      this is because of how the initial value (`null`) of the
-    //      `DiffixLowCount` aggregator is translated to `Boolean true`
     let isStarBucketLowCount =
       starBucket.Aggregators.[lowCountIndex].Final(executionContext)
       |> Value.unwrapBoolean
```
edongashi, line 201: Does `isStarBucketLowCount` depend on `bucketCount`? There's a chance that a single suppressed bucket can become high count due to the different seed.

edongashi: I see that only anon count is exported. Ok in that case.

anonymizer/OpenDiffix.Service/AggregationHooks.fs
```diff
@@ -189,20 +190,23 @@ module StarBucket =
...
 
+    let isStarBucketSingleBucket = bucketCount < 2
```
edongashi, line 203: Can also be zero buckets.

anonymizer/OpenDiffix.Service/AggregationHooks.fs
pdobacz: it doesn't.

`bucketCount` is "count of buckets in star bucket", not related to `low count filtering` or any other notion of "count" we have.

I tried to find a nicer name for this var, but couldn't so I kept it simple.

> There's a chance that a single suppressed bucket can become high count due to the different seed.

this is exactly why we're introducing the isStarBucketSingleBucket condition. If it's `true` it prevents such an "un-suppressed" bucket to surface.

edongashi (dismissed):

anonymizer/OpenDiffix.Service/AggregationHooks.fs
pdobacz: yeah, I'll switch to `bucketCount = 1`, no point in making this inclusive

anonymizer/OpenDiffix.Service/AggregationHooks.fs
edongashi: How about simplifying to:

```
    let isStarBucketLowCount =
      (if bucketCount < 2 then
        Boolean true
      else
        starBucket.Aggregators.[lowCountIndex].Final(executionContext))
      |> Value.unwrapBoolean
```

and removing the extra flag?

edongashi: We derive state closer to the source.

pdobacz: I would disagree. `bucketCount < 2` isn't exactly `star bucket low count`, it's a slightly different condition.

I'm actually thinking (I've already done it, now I'm thinking if it's correct), on rewriting it to a sequence of steps like:
1/ filter `buckets` seq, to only have those which satisfy the criteria (low count & not merged)
2/ make a second pass through the filtered seq to actually perform the mergeing.

Then the filtered seq becomes a very natural source of truth for the needed condition

pdobacz: PTAL, when you're feeling better https://github.com/diffix/desktop/pull/295/commits/64692ebcaf21b9b2e0f845e0b1be9a2691587e94

edongashi: > I would disagree. bucketCount < 2 isn't exactly star bucket low count, it's a slightly different condition.

Variable can be renamed to something else, such as `starBucketExists`.

With multiple iteration we replay the sequence which is something I'd avoid.

anonymizer/OpenDiffix.Service/AggregationHooks.fs
edongashi: If we do the single variable thing suggested above we avoid this fourth state.

edongashi (approved):

edongashi (approved):

PR #296, opened 2021-12-14
Document and explain the suppress bin

pdobacz:
Adds docs to the side panel and markdown.

I also took a stab at adding a tooltip to the suppress bin "*"-values in the table. I didn't like the result - the code was ugly and it also didn't behave well (one needs to hover exactly over the * to display the tooltip). So I came up with this dumber but straightforward way:

Screenshot from 2021-12-14 09-42-02

### DIFF
--- a/docs/operation.md
+++ b/docs/operation.md
@@ -219,7 +219,7 @@ The combined view lets you examine precisely the distortion and suppression. The
 
 ### Suppress bin
 
-The first row in the Anonymized Data table might contain the _suppress bin_, which provides the anonymized count of the data from the bins which got suppressed. The suppress bin has all of the columns selected for anonymization displaying the value `*`.
+The first row in the Anonymized Data table might contain the _suppress bin_, which provides the combined anonymized count of all the bins that were suppressed. The suppress bin shows all column values as `*`.
 
 Note that the suppress bin may itself be suppressed. The suppress bin is only shown in the `Anonymized` view.
 
--- a/docs/operation.md
+++ b/docs/operation.md
@@ -217,6 +217,12 @@ The combined view displays suppressed bins. The `Count (anonymized)` column is l
 
 The combined view lets you examine precisely the distortion and suppression. The anonymized data can be sorted by each column. Sorting by `Count (original)` shows the ascending suppressed bins first, and so is useful for examining what has been suppressed. Sorting by `Distortion` descending shows the bins with the most error first, and so is useful for examining which bins have high distortion.
 
+### Suppress bin
+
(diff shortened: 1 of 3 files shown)


src/Notebook/notebook-help.tsx
```diff
@@ -130,7 +138,10 @@ function CsvExportHelp() {
   return (
     <div>
       <Title level={4}>CSV Export</Title>
-      <Paragraph>Exports the anonymized results only (not the combined view).</Paragraph>
+      <Paragraph>
+        Exports the anonymized results only (not the combined view). The suppress bin is exported as first row in the
```
edongashi, line 142: `as the first row`?

docs/operation.md
```diff
@@ -217,6 +217,12 @@ The combined view displays suppressed bins. The `Count (anonymized)` column is l
 
 The combined view lets you examine precisely the distortion and suppression. The anonymized data can be sorted by each column. Sorting by `Count (original)` shows the ascending suppressed bins first, and so is useful for examining what has been suppressed. Sorting by `Distortion` descending shows the bins with the most error first, and so is useful for examining which bins have high distortion.
 
+### Suppress bin
+
+The first row in the Anonymized Data table might contain the _suppress bin_, which provides the anonymized count of the data from the bins which got suppressed. The suppress bin has all of the columns selected for anonymization displaying the value `*`.
```
edongashi, line 222: Some ideas how to make this shorter (feel free to disagree, I find it difficult to find the right words):

The first row in the Anonymized Data table might contain the _suppress bin_, which provides the total anonymized count of all bins that were suppressed. The suppress bin shows all column values as `*`.

I'm between `total anonymized count` or `combined anonymized count`, or any other ideas?

edongashi (approved):

src/Notebook/notebook-help.tsx
```diff
@@ -122,6 +122,15 @@ function AnonymizedResultsHelp() {
           Click here for details.
         </DocsLink>
       </Paragraph>
+      <Paragraph>
+        The{' '}
+        <DocsLink page="operation" section="suppress-bin">
+          {' '}
+          suppress bin
+        </DocsLink>{' '}
+        denoted by <Text code>*</Text> column values, is shown first. It indicates the total anonymized count of all
```
edongashi, line 131: In docs we say combined, here we say total (my mistske from the suggedtions). Maybe worth making consistent so that we don't confuse users?

edongashi (dismissed):

edongashi (approved):

PR #299, opened 2021-12-21
Rewrite and simplify LED 

edongashi:
After many attempts with various tricks, performance is finally good and reliable.

Removed heuristics and went with an exact approach. The trick was the `n` caches for `n-1` cols. Takes a little longer to build but has much more predictable runtime.

### DIFF
--- a/anonymizer/OpenDiffix.Service/Led.fs
+++ b/anonymizer/OpenDiffix.Service/Led.fs
@@ -95,6 +95,7 @@ let private prepareBuckets (aggregationContext: AggregationContext) (buckets: Bu
   let caches =
     Array.init groupingLabelsLength (fun _i -> Dictionary<Row, MutableList<Bucket * bool>>(Row.equalityComparer))
 
+  // Filters low count buckets and builds caches in the same pass.
   let victimBuckets =
     buckets
     |> Array.choose (fun bucket ->
@@ -139,24 +140,20 @@ let private led (aggregationContext: AggregationContext) (buckets: Bucket array)
       let cacheKey = cacheKeys.[colIndex]
       let siblings = caches.[colIndex].[cacheKey]
 
-      if siblings.Count = 1 then
-        // No siblings for this column (Count=1 means itself)
+      match siblings.Count with
+      | 1 ->
+        // No siblings for this column (Count=1 means itself).
+        // A column without siblings is an unknown column.
         hasUnknownColumn <- true
-
-      if siblings.Count = 2 then
-        // Single sibling. Find it and check if it's high count.
-        let siblingBucket =
-          siblings
-          |> Seq.tryPick (fun (siblingBucket, siblingLowCount) ->
-            if not (obj.ReferenceEquals(victimBucket, siblingBucket)) && not siblingLowCount then
-              Some siblingBucket
-            else
-              None
-          )
-
-        match siblingBucket with
-        | Some siblingBucket -> mergeTargets.Add(siblingBucket)
-        | None -> ()
+      | 2 ->
+        // Single sibling (self+other). Find it and check if it's high count.
+        // A column with a single high-count sibling is an isolating column.
+        for siblingBucket, siblingLowCount in siblings do
+          if not (obj.ReferenceEquals(victimBucket, siblingBucket)) && not siblingLowCount then
+            mergeTargets.Add(siblingBucket)
+      | _ ->
+        // Count=3 means there are multiple siblings and has no special meaning.
+        ()
 
     if hasUnknownColumn then diagnostics.IncrementBucketsUnknownColumn()
     if mergeTargets.Count > 0 then diagnostics.IncrementBucketsIsolatingColumn()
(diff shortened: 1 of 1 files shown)


anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -43,244 +28,174 @@ let private mergeGivenAggregatorsInto (targetBucket: Bucket) aggIndexes (sourceB
...
+
+  // For each column, we build a cache where we group buckets by their labels EXCLUDING that column.
+  // This means that every cache will associate siblings where the respective column is different.
+  let caches =
+    Array.init groupingLabelsLength (fun _i -> Dictionary<Row, MutableList<Bucket * bool>>(Row.equalityComparer))
+
+  let victimBuckets =
+    buckets
+    |> Array.choose (fun bucket ->
+      let lowCount = Utils.isLowCount lowCountIndex bucket
```
pdobacz, line 101: I needed to spend some time grokking this, so let me rephrase what it does, and you tell me if I'm on the track:

We have a single pass through buckets here, but we do two fairly separate things:
1/ choose lowCount buckets to become victimBuckets
2/ populate a "two tier" mapping <column index> -> <bucket values excluding that index> -> <that bucket>. So we effectively cache which buckets would be potential siblings for a given "n-1 element subrows".

If that's more or less the case, maybe there's a way to make it a "two stage but effectively still single pass" like

```
buckets
|> Seq.map(fun which populates caches)
|> Array.choose(fun which chooses lowCount bucket tuples)
```

If those two "tasks" are truly independent, like it seems to me, then this would read easier, and you could provide one-liner comments as to what does each step do, this would make it even easier.

You'd need to carry over the `cacheKeys` from 1st stage to the 2nd, but maybe it's not a big deal?

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -43,244 +28,174 @@ let private mergeGivenAggregatorsInto (targetBucket: Bucket) aggIndexes (sourceB
...
+  /// and has merge targets (siblings which match isolating column criteria).
+  let chooseMergeTargets (victimBucket: Bucket, cacheKeys: CacheKey array) =
+    let mutable hasUnknownColumn = false
+    let mergeTargets = MutableList()
+
+    for colIndex in 0 .. groupingLabelsLength - 1 do
+      let cacheKey = cacheKeys.[colIndex]
+      let siblings = caches.[colIndex].[cacheKey]
+
+      if siblings.Count = 1 then
```
pdobacz, line 142: so `siblings` includes "self"? I'm not sure I'm following this part. Is this dependent solely on the overall number of columns?

anonymizer/OpenDiffix.Service/Led.fs
```diff
@@ -43,244 +28,174 @@ let private mergeGivenAggregatorsInto (targetBucket: Bucket) aggIndexes (sourceB
...
+        // No siblings for this column (Count=1 means itself)
+        hasUnknownColumn <- true
+
+      if siblings.Count = 2 then
```
pdobacz, line 146: hm, how about turning this double `if` into a `match with`, where by we explicitly lay out the implications of how many siblings there were, like so:

```
| 1 -> ...
| 2 -> ...
| 3 -> //do nothing here but explain why
| _ -> failwith "some explanation why this is impossible"
```
?

Asking for this, because I am still unsure I navigate this correctly, the added prose here would probably dispel the doubts

pdobacz (commented):
ok, I think I'm getting this, but it takes some time to get through. I left some requests for comments or clarifying refactors, but overall the pieces start to fit together.

cristianberneanu:
> Removed heuristics and went with an exact approach. 

Great 👍 ! As long as the system is usable, this is always preferable. Heuristics make understanding the failure points of the mechanism very painful.

anonymizer/OpenDiffix.Service/Led.fs
edongashi: > so siblings includes "self"?

Yes, self will always end up in that slot. I had a comment somewhere for this but must have deleted it while refactoring.

anonymizer/OpenDiffix.Service/Led.fs
edongashi: This loop is the most expensive part of the hook. I experimented with parallelization but gains are minimal since there will be contention for dictionary writes (which are not thread safe). Switching to `ConcurrentDictionary` didn't help either.

I agree that the function body is a little dirty with multiple side effects, but I'd favor performance over looks since this is a hot spot.

pdobacz: but would the split as suggested hinder performance? I see that the reason for a single pass is performance, but I think the split would keep the same complexity. Hm maybe the loss would lie with the extra `cacheKeys` passed around.

So nevermind then. I'm out of ideas how to make it easier to grasp :(.

anonymizer/OpenDiffix.Service/Led.fs
edongashi: Yeah, looks better. Fixed.

pdobacz (approved):

anonymizer/OpenDiffix.Service/Led.fs
edongashi: Hah, I am about to make it more complex! I have some ideas with parallelization. If they don't work, I will try to simplify.

Note that there is already a bit of simplification because we no longer return cache keys but just the reference to the (mutable) list of siblings.

edongashi:
Ready for review

pdobacz (approved):

PR #303, opened 2022-01-03
Make renderers aware of the special case of a *-bucket value

### LINKED ISSUE #301: Generalized star bucket tooltip looks weird
image

pdobacz:
Closes #301 

I had a fix implemented already, as it is the alternative *-bucket tooltip code. Instead of manipulating the *-bucket strings by hand, we make the code aware of the special nature of the *-bucket cell value.

~I rejected this version earlier, but now, it seems much more appealing. However, I'm not entirely sure I like this code, let me know if it is too bad, I'll look for alternatives.~

EDIT: the above line and the original screenshot were outdated. Here's the behavior that this PR provides now (same as before, just the tooltip is fixed): 
Screenshot from 2022-01-04 09-53-07

### DIFF
--- a/src/AnonymizationStep/AnonymizedResultsTable.tsx
+++ b/src/AnonymizationStep/AnonymizedResultsTable.tsx
@@ -64,7 +64,7 @@ function renderValue(v: Value) {
   }
 }
 
-function renderBucketValue(column: AnonymizedResultColumn, bucketColumns: BucketColumn[]) {
+function renderBucketColumnValue(column: AnonymizedResultColumn, bucketColumns: BucketColumn[]) {
   const bucketColumn = bucketColumns.find((c) => c.name === column.name);
   return (v: Value, row: TableRowData) => {
     // see note on `ellipsis` above
@@ -127,7 +127,7 @@ const mapColumn =
       }
     }
 
-    return [makeColumnData(column.name, columnIdx, column.type, renderBucketValue(column, bucketColumns))];
+    return [makeColumnData(column.name, columnIdx, column.type, renderBucketColumnValue(column, bucketColumns))];
   };
 
 // Rows
--- a/src/AnonymizationStep/AnonymizedResultsTable.tsx
+++ b/src/AnonymizationStep/AnonymizedResultsTable.tsx
@@ -56,19 +56,6 @@ function numericRangeCell(binSize: number, v: number) {
   return <Tooltip title={tooltip}>{v}</Tooltip>;
 }
 
-function valueCell(bucketColumn: BucketColumn | undefined, v: number | boolean | string, isStarBucketRow: boolean) {
-  if (
-    bucketColumn &&
-    (bucketColumn.type === 'integer' || bucketColumn.type === 'real') &&
-    bucketColumn.generalization &&
-    !isStarBucketRow
-  ) {
-    return numericRangeCell(bucketColumn.generalization.binSize, v as number);
-  } else {
-    return plainStringCell(v.toString());
-  }
-}
-
 function renderValue(v: Value) {
   if (v === null) {
     return nullCell();
@@ -77,14 +64,23 @@ function renderValue(v: Value) {
   }
 }
 
-function buildCellRenderer(column: AnonymizedResultColumn, bucketColumns: BucketColumn[]) {
+function renderBucketValue(column: AnonymizedResultColumn, bucketColumns: BucketColumn[]) {
   const bucketColumn = bucketColumns.find((c) => c.name === column.name);
   return (v: Value, row: TableRowData) => {
     // see note on `ellipsis` above
     if (v === null) {
       return nullCell();
     } else {
-      return valueCell(bucketColumn, v, row.isStarBucketRow);
+      if (
+        bucketColumn &&
+        (bucketColumn.type === 'integer' || bucketColumn.type === 'real') &&
+        bucketColumn.generalization &&
+        !row.isStarBucketRow
+      ) {
+        return numericRangeCell(bucketColumn.generalization.binSize, v as number);
+      } else {
+        return plainStringCell(v.toString());
+      }
     }
   };
 }
@@ -131,7 +127,7 @@ const mapColumn =
       }
     }
 
-    return [makeColumnData(column.name, columnIdx, column.type, buildCellRenderer(column, bucketColumns))];
+    return [makeColumnData(column.name, columnIdx, column.type, renderBucketValue(column, bucketColumns))];
   };
 
 // Rows
--- a/src/AnonymizationStep/AnonymizedResultsTable.tsx
+++ b/src/AnonymizationStep/AnonymizedResultsTable.tsx
@@ -56,11 +56,12 @@ function numericRangeCell(binSize: number, v: number) {
   return <Tooltip title={tooltip}>{v}</Tooltip>;
 }
 
-function valueCell(bucketColumn: BucketColumn | undefined, v: number | boolean | string) {
+function valueCell(bucketColumn: BucketColumn | undefined, v: number | boolean | string, isStarBucketRow: boolean) {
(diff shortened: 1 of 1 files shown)


src/types/index.ts
```diff
@@ -113,6 +113,7 @@ export type AnonymizedValue = Value | AnonymizedAggregate;
 export type AnonymizedAggregate = {
   realValue: number;
   anonValue: number | null;
+  aggregateName: string;
```
pdobacz, line 116: I decided to add this, to not rely only on `typeof v === 'object'` given we now have 2 different kinds of objects populating the rows

edongashi (changes requested):
There is a much simpler fix. The `render` prop in [table columns](https://ant.design/components/table/#Column) accepts a row as second parameter. We can check if that row is a star bucket and return the special tooltip value. Something like this:

`AnonymizedResultsTable.tsx:L101`

```ts
const wrapRenderer = (render: (v: Value) => React.ReactNode) => (v: Value, row: TableRowData) => {
  if (row.isStarBucketRow) {
    return <Tooltip title="Suppress bin">*</Tooltip>
  } else {
    return render(v)
  }
}

function makeColumnData(
  title: string,
  dataIndex: RowDataIndex,
  type: ColumnType,
  render: (v: Value) => React.ReactNode
) {
  return {
    title,
    dataIndex,
    render: typeof dataIndex === 'string' ? render : wrapRenderer(render),
    sorter: anonymizedColumnSorter(type, dataIndex),
    ellipsis: { showTitle: false },
  };
}
``` 

Also, there's a spaghetti of `renderSomething` functions in that file which I find confusing.

pdobacz:
> There is a much simpler fix. The `render` prop in [table columns](https://ant.design/components/table/#Column) accepts a row as second parameter.

oh yeah, this is much better

edongashi:
`buildCellRenderer` already does some work so I guess it should go there. We don't need `aggregate` anyway.

pdobacz:
OK, so one commit now is a total revamp of the fix. 

>Also, there's a spaghetti of renderSomething functions in that file which I find confusing.

The second commit is my attempt to help with the spaghetti. I tried rearranging the functions but it came out much worse off. I opt to only remove one of the functions which is "off the pattern" and confusing to me and also rename one thing to be simpler.

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -76,14 +64,23 @@ function renderValue(v: Value) {
...
     // see note on `ellipsis` above
     if (v === null) {
       return nullCell();
     } else {
-      return valueCell(bucketColumn, v);
+      if (
+        bucketColumn &&
+        (bucketColumn.type === 'integer' || bucketColumn.type === 'real') &&
+        bucketColumn.generalization &&
+        !row.isStarBucketRow
```
edongashi, line 78: If we want the better tooltip we need this to be a case in itself.

src/AnonymizationStep/AnonymizedResultsTable.tsx
```diff
@@ -76,14 +64,23 @@ function renderValue(v: Value) {
   }
 }
 
-function buildCellRenderer(column: AnonymizedResultColumn, bucketColumns: BucketColumn[]) {
+function renderBucketValue(column: AnonymizedResultColumn, bucketColumns: BucketColumn[]) {
```
edongashi, line 67: This actually builds the renderer instead of rendering the value. Previous name was more honest.

src/AnonymizationStep/AnonymizedResultsTable.tsx
edongashi: Maybe `groupingLabelRenderer` is more appropriate.

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: I know, but it just didn't fit with the rest in my opinion. Every time I used to read this, I stopped to think, where it really belongs (despite I wrote the code and named the function originally).

As it is now, it communicates clearly, that it is a sibling of the `renderXYZ` functions, it's just parametrized

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: > Maybe groupingLabelRenderer is a more appropriate name.

hm, as mentioned above, I'd favor simplicity and consistency of the entire context over a strictly accurate name. Let me know how this works for you.

src/AnonymizationStep/AnonymizedResultsTable.tsx
edongashi: Ah so you mean F# style currying? Haven't been used to that in js and thought you made a mistake.

src/AnonymizationStep/AnonymizedResultsTable.tsx
edongashi: Sorry, posted in the wrong thread from phone. As for the first comment, we need to return `<Tooltip title="Suppress bin">*</Tooltip>` in a separate case, otherwise we return `<Tooltip title="*>*</Tooltip>

src/AnonymizationStep/AnonymizedResultsTable.tsx
edongashi: We have 2 competing terms for "expressions by which we have grouped by". A search for `[Gg]roupingLabel` returns 18 hits, `[Bb]ucketLabel` 3 hits, and `[Bb]ucketValue` 0 hits. Any thoughts?

edongashi: Looks like the discrepancy exists between frontend and anonymizer. If we stick to that, The proper name for this should be `renderBucketColumn`.

edongashi (dismissed):

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: > Ah so you mean F# style currying?

heh, I wasn't trying to make it that smart, only to appear similar to the other 3, but there's something to it.

> Looks like the discrepancy exists between frontend and anonymizer. If we stick to that...

yup, so this function already talks about `bucketColumns` so I thought `bucketValue` fits in nicely. 

> The proper name for this should be renderBucketColumn.

um, hold on :), we're rendering just one value in a column, not the entire column, maybe you've meant `renderBucketColumnValue` instead (which works for me)? I took `bucketValue` out of the paper BTW, but that's yet another story.

src/AnonymizationStep/AnonymizedResultsTable.tsx
pdobacz: re first comment:

It works fine, as it is yet again back to the previous behavior (which I find marginally nicer, so I kept it), as contrasted with the one I mentioned in the PR description.

Currently the *-bucket row renders as before the PR: 
Screenshot from 2022-01-04 09-53-07

Sorry about the confusion

src/AnonymizationStep/AnonymizedResultsTable.tsx
edongashi: > we're rendering just one value in a column, not the entire column

By "column" we don't really know if we're referring to the ant table column or to the SQL column (grouping label)... I meant the latter.

`bucketColumnValue` sounds good because it makes it clearer. I'm still a little unhappy that we have diverged in terminology, but this isn't the place to fix it.

src/AnonymizationStep/AnonymizedResultsTable.tsx
edongashi: Ah, I thought the new behavior was desired.

edongashi (approved):

PR #307, opened 2022-01-10
Tweak anonymization configuration section title.

### LINKED ISSUE #285: Change "Adjust suppression threshold" to just "Suppression threshold"
The default of 3 normally doesn't have to be adjusted, so I'd like to remove "Adjust" from the heading so that there is no implication that it needs to be done.

cristianberneanu:
Closes #285.

### DIFF
--- a/src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
+++ b/src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
@@ -33,7 +33,7 @@ export const AnonParamsSelectionStep: FunctionComponent<AnonParamsSelectionStepP
     <>
       <div className="AnonParamsSelectionStep notebook-step">
         <NotebookNavAnchor step={NotebookNavStep.AnonParamsSelection} status="done" />
-        <Title level={3}>Anonymization parameters</Title>
+        <Title level={3}>Suppression configuration</Title>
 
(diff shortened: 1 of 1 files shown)


src/AnonParamsSelectionStep/AnonParamsSelectionStep.tsx
```diff
@@ -33,7 +33,7 @@ export const AnonParamsSelectionStep: FunctionComponent<AnonParamsSelectionStepP
     <>
       <div className="AnonParamsSelectionStep notebook-step">
         <NotebookNavAnchor step={NotebookNavStep.AnonParamsSelection} status="done" />
-        <Title level={3}>Adjust suppression threshold</Title>
+        <Title level={3}>Anonymization parameters</Title>
```
edongashi, line 36: Paul made a related change today:

https://github.com/diffix/desktop/blob/bbe74a7e33e923d884ba396ae43d7f7d6800baa2/src/Notebook/notebook-nav.tsx#L286-L290

Should these be in sync?

cristianberneanu: Ah, missed that. I will make the section heading similar to the step title.

pdobacz (approved):

PR #308, opened 2022-01-11
Fix crash when exporting global count.

cristianberneanu:
Star bucket hook requires a present LC aggregator, which is not available when exporting the global count.

### DIFF
--- a/anonymizer/OpenDiffix.Service/Program.fs
+++ b/anonymizer/OpenDiffix.Service/Program.fs
@@ -190,10 +190,17 @@ let handleExport
 
   let countColumn = $"diffix_count(%s{countInput}, %s{aidColumn}) AS count"
 
-  let query =
+  // We get a hold of the star bucket results reference via side effects.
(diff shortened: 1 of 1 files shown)


edongashi (approved):
Sorry, missed the assignment.

anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -190,27 +190,25 @@ let handleExport
 
   let countColumn = $"diffix_count(%s{countInput}, %s{aidColumn}) AS count"
 
-  let query =
+  // We get a hold of the star bucket results reference via side effects.
+  let mutable suppressedAnonCount = Null
```
edongashi, line 194: The hook should be smart enough to ignore global queries, like Led does.

cristianberneanu: The problem is the missing mandatory aggregator. 
I thought it is simpler to conditionally call the hook, rather than modify the hook to handle such cases.

PR #322, opened 2022-01-18
Fix LCF threshold mean to be LowMeanGap*SD away from LowThreshold

pdobacz:


pdobacz:
To my surprise, this failed tests. This is because the ` * sqrt(2)` made the suppression threshold effectively higher, so our sample data (and sample Export request) didn't contain any non-suppressed bins :(.

This is the second case of these tests proving out brittle, so I'll make a self-note to figure out something better in their place.

cristianberneanu (approved):

Issue #282, opened 2021-12-06
Always cast columns to their inferred type when loading from the table


edongashi:
We have to normalize column values in `mapBucketToSQL` to their proper type so that grouping and hooks get to work with proper types instead of strings. 

https://github.com/diffix/desktop/blob/fb318141d2d1c7e76fb8ed8f2c38744e94e7400c/src/shared/anonymizer.ts#L63-L79

PR #327, opened 2022-03-02
Test the CSV exporting capabilities separately - stabilize Export tests

pdobacz:
A little thing I put off for a longer while - the original `Program` tests for service proved brittle twice - they relied on the random data and the suppression params just to test the functionalities of CSV formatting. This quick PR fixes it by introducing a dedicated module and testing it separately.

### DIFF
--- a/anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
+++ b/anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
@@ -1,0 +1,82 @@
+module OpenDiffix.Service.CSVFormatterTests
+
+open Xunit
+open FsUnit.Xunit
+
+open OpenDiffix.Core
+open OpenDiffix.Core.QueryEngine
+
+[<Fact>]
+let ``Formats result with star bucket`` () =
+  let result =
+    {
+      Columns =
+        [
+          { Name = "age"; Type = IntegerType }
+          { Name = "city"; Type = StringType }
+          { Name = "discount"; Type = RealType }
+          { Name = "active"; Type = BooleanType }
+          { Name = "count"; Type = IntegerType }
+        ]
+      Rows = [ [| Integer 18L; String "Berlin"; Real 0.5; Boolean true; Integer 1L |] ]
+    }
+
+  let suppressedAnonCount = Integer 2L
+  let CSVResult = (CSVFormatter.format suppressedAnonCount result).Split('\n')
+
+  CSVResult.[0]
+  |> should equal "\"age\",\"city\",\"discount\",\"active\",\"count\""
+
+  CSVResult.[1] |> should equal "\"*\",\"*\",\"*\",\"*\",2"
+  CSVResult.[2] |> should equal "18,\"Berlin\",0.5,True,1"
+  CSVResult.Length |> should equal 3
+
+[<Fact>]
+let ``Formats result without the star bucket`` () =
+  let result =
+    {
+      Columns = [ { Name = "count"; Type = IntegerType } ]
+      Rows = [ [| Integer 1L |] ]
+    }
+
+  let suppressedAnonCount = Null
+  let CSVResult = (CSVFormatter.format suppressedAnonCount result).Split('\n')
+
+  CSVResult.[0] |> should equal "\"count\""
+  CSVResult.[1] |> should equal "1"
+  CSVResult.Length |> should equal 2
+
+[<Fact>]
+let ``Handles quotes`` () =
+  let result =
+    {
+      Columns = [ { Name = "name"; Type = StringType } ]
+      Rows = [ [| String "\"John Quoted\"" |] ]
+    }
+
+  let suppressedAnonCount = Null
+  let CSVResult = (CSVFormatter.format suppressedAnonCount result).Split('\n')
+
+  CSVResult.[1] |> should equal "\"\"\"John Quoted\"\"\""
+  CSVResult.Length |> should equal 2
+
+[<Fact>]
+let ``Formats nulls`` () =
+  let result =
+    {
+      Columns =
+        [
+          { Name = "age"; Type = IntegerType }
+          { Name = "city"; Type = StringType }
+          { Name = "discount"; Type = RealType }
+          { Name = "active"; Type = BooleanType }
+          { Name = "count"; Type = IntegerType }
+        ]
+      Rows = [ [| Null; Null; Null; Null; Null |] ]
+    }
+
+  let suppressedAnonCount = Null
+  let CSVResult = (CSVFormatter.format suppressedAnonCount result).Split('\n')
+
+  CSVResult.[1] |> should equal "NULL,NULL,NULL,NULL,NULL"
+  CSVResult.Length |> should equal 2
--- a/anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
+++ b/anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
@@ -7,7 +7,7 @@ open OpenDiffix.Core
 open OpenDiffix.Core.QueryEngine
 
 [<Fact>]
-let ``Formats result with star bucket`` () =
+let ``Formats result`` () =
   let result =
     {
       Columns =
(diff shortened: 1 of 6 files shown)


anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
```diff
@@ -0,0 +1,82 @@
...
+          { Name = "active"; Type = BooleanType }
+          { Name = "count"; Type = IntegerType }
+        ]
+      Rows = [ [| Null; Null; Null; Null; Null |] ]
+    }
+
+  let suppressedAnonCount = Null
+  let CSVResult = (CSVFormatter.format suppressedAnonCount result).Split('\n')
+
+  CSVResult.[1] |> should equal "NULL,NULL,NULL,NULL,NULL"
```
edongashi, line 81: Aren't we supposed to treat `NULL` as empty string?

pdobacz: good question, but it seems it's been `NULL` for a long time, c.f. 
https://github.com/diffix/desktop/blob/master/anonymizer/OpenDiffix.Service/Program.fs#L36
and https://github.com/diffix/reference/blob/master/src/OpenDiffix.Core/Value.fs#L14

edongashi: That looks like a pseudo-csv that we have been emitting. We may need to represent `NULL` as `""`. @cristianberneanu?

anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
```diff
@@ -0,0 +1,82 @@
...
+  let result =
+    {
+      Columns = [ { Name = "name"; Type = StringType } ]
+      Rows = [ [| String "\"John Quoted\"" |] ]
+    }
+
+  let suppressedAnonCount = Null
+  let CSVResult = (CSVFormatter.format suppressedAnonCount result).Split('\n')
+
+  CSVResult.[1] |> should equal "\"\"\"John Quoted\"\"\""
```
pdobacz, line 60: Actually, I was supposed to double check @edongashi - are we indeed triple quoting in this case? This doesn't look right either

anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
cristianberneanu: > when I plant some NULLs (,,) into the input CSV

CSV format doesn't support NULL fields. The parser probably converts those to "" automatically.
The Desktop export query should never generate NULL values in the output, so we shouldn't worry about it.

pdobacz: Sorry about raising false alert on this, and thank you for raising the issue. I'm scratching my head here - if I inject a `printfn` into `CSVFormatter` result and run it, I'm getting different results in the `printfn`, depending on how I run it:
1/ `NULL` if I run the `CSVFormatterTests`
2/ `""` if I run `OpenDiffix.Service` `Export` request manually

I'm going to investigate why this happens, I'm sure the problem lies with the tests, but in some obscure manner.

anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+  let suppressedAnonCount = Integer 2L
+  let CSVResult = (CSVFormatter.format suppressedAnonCount result).Split('\n')
+
+  CSVResult.[0]
+  |> should equal "\"age\",\"city\",\"discount\",\"active\",\"count\""
+
+  CSVResult.[1] |> should equal "\"*\",\"*\",\"*\",\"*\",2"
+  CSVResult.[2] |> should equal "18,\"Berlin\",0.5,True,1"
```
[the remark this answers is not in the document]
pdobacz: OK, I think mystery solved. The import part interprets the data as `passenger_count` being a string, this is why `desktop` seemed to behave correctly. Funny part is that I can generalize it, and then the output column is integer-typed, giving the `NULL` cell out in the CSV.

```
"passenger_count","count"
NULL,11
0,324315
2,67317
4,33552
6,14969
```

I'll hunt this down.

anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
edongashi: Like Cristian said. The output of the csv reader has no nulls, whereas your tests may have them specifically set.

anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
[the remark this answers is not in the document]
edongashi: Yes, this is addressed in #282.

anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
pdobacz: I am able to export NULLs from `desktop v1.0.1` if I use generalization on the integer column. Also, even without the nulls in the input data, we provide inconsistent output data in generalized / non-generalized request:

```
"passenger_count","count"
2,67315
0,324384
4,33573
6,15004
```

```
"passenger_count","count"
"2",52441
"1",324332
"4",6345
"3",14917
"5",27193
"6",15007
```

Generalization causes the query to `cast(passenger_count as real)` which removes the quote and enables the export to produce `NULL`, if any are present in the input

anonymizer/OpenDiffix.Service.Tests/CSVFormatter.Tests.fs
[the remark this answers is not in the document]
edongashi:
```
"passenger_count","count"
NULL,11
```

Seems that we may need to coerce to empty string after all...

anonymizer/OpenDiffix.Service/CSVFormatter.fs
```diff
@@ -0,0 +1,38 @@
...
+
+let private quoteString (string: string) =
+  "\"" + string.Replace("\"", "\"\"") + "\""
+
+let private csvFormat value =
+  match value with
+  | String string -> quoteString string
+  | _ -> Value.toString value
+
+let format suppressedAnonCount (result: QueryResult) =
```
edongashi, line 15: It's bad abstraction to be aware of `suppressedAnonCount` here. The star bucket row can be prepended to the row sequence at `O(1)` if we allow for a `seq<Row>` input.

anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -56,6 +56,21 @@ let getAnonParams (requestAnonParams: RequestAnonParams) (salt: string) =
     LayerNoiseSD = requestAnonParams.LayerNoiseSD
   }
 
+let prependStarBucketRow suppressedAnonCount result =
+  if suppressedAnonCount <> Null then
+    let starBucketRow =
+      result.Columns
+      |> List.map (
+        function
+        | { Name = "count" } -> suppressedAnonCount
```
edongashi, line 65: What if we have a regular column named `count`?

anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -174,13 +189,12 @@ let handleExport
...
-  let output =
+  let result =
     anonParams
     |> runQuery hooks query inputPath
-    |> CSVFormatter.format suppressedAnonCount
-
-  File.WriteAllText(outputPath, output)
+    |> prependStarBucketRow suppressedAnonCount
 
+  File.WriteAllText(outputPath, CSVFormatter.format result)
```
edongashi, line 197: Thanks for the change, looks cleaner!

edongashi (approved):

anonymizer/OpenDiffix.Service/Program.fs
pdobacz: hm, true, IIRC `Service` inherits the magic meaning of `count` from `desktop`. But this is a good question to actually check in general, what would happen, I can't remember anymore. As this PR doesn't change this in particular, I'll leave this up to a follow-up (hopefully, without such delay)

PR #328, opened 2022-03-09
Fix bug with handling *-bucket and columns named `count`

pdobacz:
Found by Edon in https://github.com/diffix/desktop/pull/327#discussion_r818406789. The handling in the Electron side of things (`Preview` requests) is fine, only `Export` requests were broken, contrary to what I posted there.

I admit, I had a hard time trying to make the code arranging columns and indices foolproof and readable. Unless you have any ideas to simplify this easily, it seems like making this code nice would require a larger uprooting (like making `QueryEngine` aware of what the columns it outputs represents (labels vs aggregators)... but this doesn't sound right).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### Next version
+
+- Fixed input data with a "count" column resulting in incorrect suppress bin.
+
 ### Version 1.0.1
 
 - Adjusts the computation of the suppression threshold mean to spec.
--- a/anonymizer/OpenDiffix.Service/Program.fs
+++ b/anonymizer/OpenDiffix.Service/Program.fs
@@ -56,15 +56,11 @@ let getAnonParams (requestAnonParams: RequestAnonParams) (salt: string) =
     LayerNoiseSD = requestAnonParams.LayerNoiseSD
   }
 
-let prependStarBucketRow suppressedAnonCount result =
+let prependStarBucketRow suppressedAnonCount countColumnIndex result =
   if suppressedAnonCount <> Null then
     let starBucketRow =
       result.Columns
-      |> List.map (
-        function
-        | { Name = "count" } -> suppressedAnonCount
-        | _ -> String "*"
-      )
+      |> List.mapi (fun columnIdx _ -> if columnIdx = countColumnIndex then suppressedAnonCount else String "*")
       |> Array.ofList
 
     { result with Rows = starBucketRow :: result.Rows }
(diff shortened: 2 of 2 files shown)


anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -56,15 +56,11 @@ let getAnonParams (requestAnonParams: RequestAnonParams) (salt: string) =
...
+let prependStarBucketRow suppressedAnonCount countColumnIndex result =
   if suppressedAnonCount <> Null then
     let starBucketRow =
       result.Columns
-      |> List.map (
-        function
-        | { Name = "count" } -> suppressedAnonCount
-        | _ -> String "*"
-      )
+      |> List.mapi (fun columnIdx _ -> if columnIdx = countColumnIndex then suppressedAnonCount else String "*")
```
cristianberneanu, line 63: Inconsistent abbreviation of index in `columnIdx` compared to the rest of the module.

anonymizer/OpenDiffix.Service/Program.fs
```diff
@@ -170,29 +166,38 @@ let handleExport
...
   let countColumn = $"diffix_count(%s{countInput}, %s{aidColumn}) AS count"
+  let columns = Array.concat [ buckets; [| countColumn |] ]
+  let countColumnIndex = Array.findIndex (fun column -> column = countColumn) columns
+
+  let bucketColumnGroupByIndices =
+    columns
+    |> Array.indexed
+    |> Array.filter (fun (_, column) -> column <> countColumn)
+    // GROUP BY expects 1-based indexing
+    |> Array.map (fst >> (+) 1)
```
cristianberneanu, line 177: I don't understand why this part is needed at all. The `SELECT` clause is manually assembled right here, so we know that the `count` column is always the last one. The old version of the code seems fine to me.

pdobacz: OK, I thought that it was easier to mix things up that way, but I wasn't sure, I can take it all back and just fix the bug surgically.

edongashi (approved):

pdobacz:
@edongashi sorry about this, I forgot to update `CHANGELOG.md`, now fixed

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 
+### Next version
+
+- Fixed input data with a "count" column resulting in incorrect suppress bin
```
edongashi, line 5:
```suggestion
- Fixed input data with a "count" column resulting in incorrect suppress bin.
```

edongashi (approved):

edongashi (approved):

CHANGELOG.md
edongashi: I updated my comment slightly. It's fine as it is. Though it sounds that you "fixed" the input data instead of the bug.

PR #329, opened 2022-03-10
Make CSVDataProvider aware of inferred column types

### LINKED ISSUE #282: Always cast columns to their inferred type when loading from the table

pdobacz:
closes #282 

In particular, this solves the problem I stumbled upon, which is that when you have an integer column in the input CSV, the type of it's bucket column, which comes out as exported from DfD depends:
1. If you don't generalize it, it will be exported as string (and surrounded by `"`s)
2. If you generalize it, it will be exported as int (without the `"`s)

I tried to follow the logic of type inference used in `desktop`, in typescript code. That is, if the column consistently follows some type, it will be that type, and we look only at the first 1000 rows, to make things manageable.

Sorry about the rough edges, but I thought I'd go ahead and PR early to get your feedback, before trying to figure out 100% myself. I'll leave some comments on what in particular is still bugging me.

### DIFF
--- a/anonymizer/OpenDiffix.Service/CSVDataProvider.fs
+++ b/anonymizer/OpenDiffix.Service/CSVDataProvider.fs
@@ -18,22 +18,123 @@ let private openCsvStream stream =
 
   csv
 
-type DataProvider(stream: TextReader) =
-  let csv = openCsvStream stream
+type private TypeInfo = { IsEmpty: bool; IsBoolean: bool; IsInteger: bool; IsReal: bool }
 
-  new(dbPath: string) = new DataProvider(new StreamReader(dbPath))
+let private unknownTypeInfo = { IsEmpty = true; IsBoolean = true; IsInteger = true; IsReal = true }
+let private integerTypeInfo = { IsEmpty = false; IsBoolean = false; IsInteger = true; IsReal = true }
+
+let private parseBoolean (field: string) =
+  if field = "1" then
+    Boolean true
+  else if field = "0" then
+    Boolean false
+  else
+    match Boolean.TryParse field with
+    | (true, value) -> Boolean value
+    | (false, _) -> Null
+
+let private parseInt64 (field: string) =
+  match Int64.TryParse field with
+  | (true, value) -> Integer value
+  | (false, _) -> Null
+
+let private parseDouble (field: string) =
+  match Double.TryParse field with
+  | (true, value) -> Real value
+  | (false, _) -> Null
+
+let private tryInferType (field: string) =
+  if field = "" then
+    Some unknownTypeInfo
+  else
+    Some
+      {
+        IsEmpty = false
+        IsBoolean = parseBoolean field <> Null
+        IsInteger = parseInt64 field <> Null
+        IsReal = parseDouble field <> Null
+      }
+
+let private deriveColumnType (columnTypeInfo: TypeInfo) =
+  if columnTypeInfo.IsEmpty then StringType
+  else if columnTypeInfo.IsBoolean then BooleanType
+  else if columnTypeInfo.IsInteger then IntegerType
+  else if columnTypeInfo.IsReal then RealType
+  else StringType
+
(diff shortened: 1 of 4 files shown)


anonymizer/OpenDiffix.Service/CSVDataProvider.fs
```diff
@@ -18,22 +18,123 @@ let private openCsvStream stream =
...
+let private parse columnType field =
+  match columnType with
+  | BooleanType -> parseBoolean field
+  | IntegerType -> parseInt64 field
+  | RealType -> parseDouble field
+  | _ -> String field
+
+type DataProvider(stream: TextReader, streamForTypes: TextReader) =
   let csv = openCsvStream stream
+  let csvForTypes = openCsvStream streamForTypes
```
pdobacz, line 100: Is there a better way to "reset" the reader. We need to scan the file twice - first on the construction (1000 rows max, for type inference) and second time on `OpenTable` which yields rows. Keeping two `StreamReader` instances was the only thing I managed to find :/.

anonymizer/OpenDiffix.Service/CSVDataProvider.fs
```diff
@@ -18,22 +18,123 @@ let private openCsvStream stream =
 
   csv
 
-type DataProvider(stream: TextReader) =
+type private TypeInfo = { IsEmpty: bool; IsBoolean: bool; IsInteger: bool; IsReal: bool }
+
+let private unknownTypeInfo = { IsEmpty = true; IsBoolean = true; IsInteger = true; IsReal = true }
+let private integerTypeInfo = { IsEmpty = false; IsBoolean = false; IsInteger = true; IsReal = true }
```
pdobacz, line 24: this is here to handle the `RowIndex` column. I tried some alternatives, but this remains as the least obtrusive one.

edongashi:
I'm not sold that it's worth doing this in the data provider. A simpler approach would be to send type info for the labels to `service`. Non-generalized columns can be cast like the following:

```sql
SELECT
  text_col, // no cast
  cast(int_col as integer) as int_col,
  cast(real_col as real) as real_col,
  count(*)
FROM table
GROUP BY 1, 2, 3;
```

With this slight tweak in emitted sql we can fix the quoting issue.

cristianberneanu:
Yes, duplicating the type detection logic in the service seems redundant. Making the type casts always present when selecting columns seems like the more direct and simple approach.

anonymizer/OpenDiffix.Service/CSVDataProvider.fs
pdobacz: Yup, this would work. The question is though: how do you think, is it worth it? It's hard for me to judge without plunging in, how involved would the cache code be in practice.

pdobacz:
> I'm not sold that it's worth doing this in the data provider...

> Yes, duplicating the type detection logic in the service seems redundant. Making the type casts always present when selecting columns seems like the more direct and simple approach...

I like this idea, but I've got some reservations - it seems a tiny bit hacky after all. It makes `service` a bit reliant on the other pieces and specifics of the query, we impose some rather unnatural additions to the query.

As far as separation of concerns, I'd say that only `service` (or wherever `CSVDataProvider` resides) should be responsible for type inference and report it back to `desktop`. 

But all that being said, what you suggest sounds like a practical thing to do right now.

pdobacz:
I guess if we document well, that "`CSVDataProvider` reads all `Value`s as `String` **by design** and the query is responsible for casting them into expected types", it would be sound. The caller of `service` is responsible for type inference, it's just the way to communicate it to `service` is slightly non-standard (via SQL casts).

PR #330, opened 2022-03-14
Force casting columns as their detected types

### LINKED ISSUE #282: Always cast columns to their inferred type when loading from the table

pdobacz:
Closes #282 

Supersedes #329

### DIFF
--- a/anonymizer/OpenDiffix.Service/CSVFormatter.fs
+++ b/anonymizer/OpenDiffix.Service/CSVFormatter.fs
@@ -9,7 +9,7 @@ let private quoteString (string: string) =
 let private csvFormat value =
   match value with
   | Null -> ""
-  // Lowercase `Boolean` in order to be consistent with JavaScript rendering.
+  // Lowercase `Boolean` in order to be consistent with the UI.
   | Boolean false -> "false"
   | Boolean true -> "true"
   | String string -> quoteString string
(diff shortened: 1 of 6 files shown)


anonymizer/OpenDiffix.Service/CSVFormatter.fs
```diff
@@ -9,6 +9,9 @@ let private quoteString (string: string) =
 let private csvFormat value =
   match value with
   | Null -> ""
+  // Lowercase `Boolean` in order to be consistent with JavaScript rendering.
```
pdobacz, line 12: I'm not happy about this, but the discrepancy (`true/false` in preview in Dfd vs `True/False` in exported CSV) was bugging me. It's slightly easier to fix this way around

edongashi: What does `JavaScript rendering` mean?

edongashi (dismissed):

anonymizer/OpenDiffix.Service/CSVFormatter.fs
pdobacz: hm, I guess what is happening when `true/false` becomes a string (`.toString()` I presume). I didn't find any code handling that in DfD, so I guess this is just the default. THat happens when we're rendering tables. I don't like the wording either, any suggestions?

PR #336, opened 2022-05-17
Fix type casting of generalized `integer` columns.

### LINKED ISSUE #332: Numeric generalization of `integer` columns casts to `real`.
This prevents expression normalization in `reference`.

cristianberneanu:
Closes #332.


pdobacz (dismissed):
LGTM, I remember this has been a pain point. Did you double check exporting works as expected with/without generalization? I think it should be fine, but not 100% sure.

cristianberneanu:
> LGTM, I remember this has been a pain point. Did you double check exporting works as expected with/without generalization? I think it should be fine, but not 100% sure.

Good point and you're right, it doesn't work  because the cast to `integer` acts as a `round`, so the expression is rejected.

pdobacz (approved):

PR #340, opened 2022-08-10
Language switching support

edongashi:
Will dump German strings once we agree on the loc file shape.


assets/locales/en/translation.json
```diff
@@ -1 +1,106 @@
...
+    "Export anonymized data to CSV": "Export anonymized data to CSV"
+  },
+  "SchemaLoadStep": {
+    "Loading schema": "Loading schema",
+    "Successfully imported {{fileName}}": "Successfully imported {{fileName}}",
+    "Here is what the data looks like:": "Here is what the data looks like:",
+    "(only the first 1000 rows are shown)": "(only the first 1000 rows are shown)",
+    "Loaded {{fileName}}": "Loaded {{fileName}}"
+  }
+}
```
cristianberneanu, line 183: Since the key equals the value for english text, can't we omit this mapping entirely and avoid updating the strings twice?

src/main.ts
```diff
@@ -67,6 +67,28 @@ function setupMenu() {
...
+            {
+              label: 'Deutsch',
+              type: 'radio',
+              checked: i18n.language === 'de',
+              click: () => i18n.changeLanguage('de'),
+            },
+          ],
+        },
+      ],
+    },
```
cristianberneanu, line 107: How about also putting a flag icon somewhere in the UI?

cristianberneanu (approved):

assets/locales/en/translation.json
edongashi: It's "Possible - but not recommended" with [key fallback](https://www.i18next.com/principles/fallback#key-fallback), but that example assumes non-scoped keys. With scoped keys (which I prefer) we need to wrap the `t` func result as well, otherwise we get `App::Documentation` as the fallback value. We also get spammed with key not found messages in debug logs. Will think about it more, but not a high priority in my opinion.

src/main.ts
edongashi: Any idea where? If we put it inside the notebook view it looks like the language change is local to that notebook only...

edongashi: Let's revisit later.

PR #344, opened 2022-09-20
Autodetect language based on app locale

### LINKED ISSUE #343: Test/support auto detection of language to use for the GUI

pdobacz:
Closes #343 .

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -196,6 +196,10 @@ function createWindow() {
 // IPC
 
 app.on('ready', () => {
+  // Feeding the locale to `changeLanguage` or extracting the language cause problems.
+  if (['de', 'de-AT', 'de-CH', 'de-DE', 'de-LI', 'de-LU'].includes(app.getLocale())) {
+    i18n.changeLanguage('de');
+  }
   setupMenu();
   registerProtocols();
   createWindow();
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ folder in this repo.
 Run `asdf install` to install `node` via `asdf`.
 Run `npm install` to install dependencies.
 Run `git submodule update --init && npm run build` to get and build the anonymizer.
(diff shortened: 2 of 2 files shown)


src/main.ts
```diff
@@ -196,6 +196,11 @@ function createWindow() {
 // IPC
 
 app.on('ready', () => {
+  // @ts-ignore
```
pdobacz, line 199: I have been struggling to get rid of this. It is here, because TS complains about "`Property 'Locale' does not exist on type 'typeof Intl'`", which makes little sense to me. Everything works fine if I ignore the error, my IDE doesn't complain. I have not been able to find any useful explanation on the web.

On the other hand the `Intl.Locale` is what I've found to reliably extract the language code from the locale code. This all feels clunky and out of place, am I missing something? This should be simple :/.

Nevertheless, works on Ubuntu and Windows, tried English, Polish, German and German-Liechtenstein localized systems, behaves as expected.

edongashi: Can we simplify to `if (['de', 'de-AT', 'de-CH', 'de-DE'].includes(app.getLocale()) { change lang ... }`?

edongashi: Is this an intrinsic v8 type, because I don't see an import anywhere?

pdobacz: ok, it turns out always worked with locale sent as language, it only caused a bunch of errors, but worked anyway :shrug:. THere's even a setting which makes these errors go away. It still complains if the language is not supported, but I think this is good enough.

edongashi: Are these errors due to the fact that we assign some unkown language, e.g. Polish, and then i18next complains about missing keys? That's not ideal. I suggest a simple check if the app locale is `de` and switching to it.

pdobacz: yyyyeah, it actually errors on each render. I give up, explicit listing of `de` locales it is...

edongashi (approved):

PR #345, opened 2022-11-12
remove `\n` row in taxi.csv

yoid2000:
I removed the `\n` row. I don't think any review is needed, just approve and merge, whoever gets to it first.


cristianberneanu:
Adding support for this NULL marker should be pretty easy and we'll have to do it anyway, since Postgres uses it when exporting to CSV. I think it is better to keep the file as it is, so we can catch any regressions for this case in the future.

yoid2000:
I made a copy of the file with the NULL markers. Once the bug is fixed, we can move the copy back. But I'd rather not have a situation where the file we offer as a sample doesn't actually work...

cristianberneanu (approved):

